PR #10, opened 2023-10-19
WIP: run a sqlite query

kmcq:
TODO:
- [ ] query one of the chinook tables and see the results
  - [x] get tanstack set up
  - [ ] something that executes the query
  - [ ] form to type query in
  - [ ] workflow that renders the data
    - shows all columns? allowlist which ones?
    - column naming patterns
      - starts with `_` --> hide
      - ends with `_url` --> linkify. assumes the value of the column is a URI or URL?


kmcq:
closing this out, we're doing other things

PR #18, opened 2023-11-02
Setup typescript

hktouw:
@open-truss/engineers

### DIFF
--- a/packages/open-truss/tsconfig.json
+++ b/packages/open-truss/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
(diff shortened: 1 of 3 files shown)


packages/open-truss/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
jonmagic, line 3: https://www.typescriptlang.org/tsconfig#target

```suggestion
    "target": "es6",
```

jonmagic (approved):
One suggestion. Where did you get the `tsconfig.json` from? Was it auto created? Just wondering if there is a best practices starting point for a tsconfig in a new project.

hktouw:
Oh, I more or less copied the one @kmcq made in the demo app https://github.com/open-truss/open-truss/blob/main/demo-app/tsconfig.json

kmcq:
> Oh, I more or less copied the one @kmcq made in the demo app https://github.com/open-truss/open-truss/blob/main/demo-app/tsconfig.json

Just to add the remaining context: this came from the create-next-app starter script

PR #19, opened 2023-11-02
Fix package publishing

hktouw:
Fixes the script that publishes a new package every time a PR merges. Can see the package successfully publish in this actions run https://github.com/open-truss/open-truss/actions/runs/6738603968/job/18318529441

One quirk that I'm not realizing is that we'll need to bump the package version on every PR. A couple thoughts:
- We could add a CI step that fails the build if the PR has a package version lower than what is on main. That will remind us to bump the version.
- If bumping the package version every time we open a PR sounds tedious, perhaps we should abandon this approach altogether. We can disable this actions job and rely on publishing packages manually for now. Eventually when this is in a stable version we can trigger publishes when we open releases in GitHub.

@open-truss/engineers which approach do you all prefer?

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -2,7 +2,7 @@ name: Publish Package to npmjs
 on:
   push:
     branches:
-      - fix-package-publishing
+      - main
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -2,7 +2,7 @@ name: Publish Package to npmjs
 on:
   push:
     branches:
-      - main
+      - fix-package-publishing
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


jonmagic:
I think it's ok for a package publish to fail if we're updating something other than the package or aren't ready to release a new version. When we are ready to publish a new version of any of the packages in the repo then we bump the version and the next merge to main publishes it. My 2 cents.

Maybe we could have it just skip package publishing if there isn't a version bump? So instead of appearing as a failed actions run it just appears as a noop actions run?

jonmagic (approved):

hktouw:
> Maybe we could have it just skip package publishing if there isn't a version bump? So instead of appearing as a failed actions run it just appears as a noop actions run?

Great idea. Just updated the PR to do this. Turns out there is a nifty action that handles this for us.

PR #20, opened 2023-11-04
Add linter

hktouw:
## Why

We want to have a linter build step generally. More importantly, it seems like at least one status check is required in order to require PRs be up to date before merging. To enable this, I went ahead and adding a linting status check.

> **Require branches to be up to date before merging**
> This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).

Settings · Branches · Branch protection rule · open-trussopen-truss 2023-11-04 at 12 15 08 PM

## How

This PR adds eslint to the project and runs it as a status check.

@open-truss/engineers

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -1,10 +1,7 @@
 run-name: CI Checks on ${{ github.head_ref}} by @${{ github.actor }}
 
 on:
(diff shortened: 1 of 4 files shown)


jonmagic (approved):
let the linter gods strike us down

PR #21, opened 2023-11-04
Re-arrange project and transpile code in build

hktouw:
## Why

We want to rearrange the project to make development easier and update the build step to transpile the code.

## How

This PR modifies the structure of the project to mirror react/primer since that is probably a reasonably well setup project. from https://github.com/open-truss/open-truss/commit/68a2a980979df16a1574c4685f9d9ee834e2c069

> This code sets up the project to be modeled after primer/react.
> Modules are made in src/{module}/index.ts and imported into
> src/index.ts. The src/index.ts file is the entry point for applications
> and exports all modules. See
>
> https://github.com/primer/react/blob/main/src
> https://github.com/primer/react/blob/main/src/index.ts

Additionally this PR adds compilation / transipilation to the build step.

The easiest way to review this PR is starting from this committ (https://github.com/open-truss/open-truss/commit/68a2a980979df16a1574c4685f9d9ee834e2c069) and ending with (https://github.com/open-truss/open-truss/pull/21/commits/b60d93a8edb9db53344a0465d2a9fc87cc45886a), reading the comments along the way.

@open-truss/engineers


jonmagic (approved):
Great work!

PR #27, opened 2023-11-08
Add `formatQuery` util

andrew-dillon:
Paired with @pourash-b @jonmagic @hktouw.

Created a `formatQuery` util function that uses [`sql-formatter`](https://github.com/sql-formatter-org/sql-formatter/) to inject named variables into a SQL template string. Supports strings, numbers, and arrays of strings/numbers as variable types. 

This should support just about any conceivable query use case. In the future there _might_ be other types to support as well, e.g. dates/times?

**Note:** It's worth calling out that these variables are not truly escaped prior to injection in the template. The only safe way to do this, to my knowledge, is to let the database engine inject the variables. But since open truss is build on the concept of running dynamic queries from trusted clients, I don't think this is a concern?


package-lock.json
jonmagic, line 1: I think this file can be removed?

packages/open-truss/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default as helloWorld } from './hello_world'
-export { parseYaml, stringifyYaml } from './utils/yaml'
+export * from './utils/yaml'
+export * from './utils/format'
```
jonmagic, line 3: Can you update your editor to add a newline at the end of a file and if you don't have it already also remove whitespace from the end of lines.

packages/open-truss/src/utils/format.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function formatQuery(template: string, config?: Config) {
+  return sqlFormat(template, {
+    language: config?.language || 'sql',
+    tabWidth: 4,
+    linesBetweenQueries: 2,
+    paramTypes: { named: [':'] },
+    params: config?.params && formatParams(config.params),
+  })
+}
```
jonmagic, line 37: Same here with the final newline.

jonmagic (approved):
A few minor nits but otherwise looks great!

Can we settle on an eslint config that applies the settings I talked about using for your editor? Ooh, maybe we can use https://editorconfig.org/?

> It's worth calling out that these variables are not truly escaped prior to injection in the template. The only safe way to do this, to my knowledge, is to let the database engine inject the variables. But since open truss is build on the concept of running dynamic queries from trusted clients, I don't think this is a concern?

Good call out and I agree with your conclusion.

andrew-dillon:
@jonmagic yeah I'm all about formatters and standardizing code style. What do you think about https://prettier.io?

jonmagic:
> @jonmagic yeah I'm all about formatters and standardizing code style. What do you think about https://prettier.io?

Honestly I'm not sure which tools we should be using to define and apply formatting rules. Do eslint and prettier work together? Should we use one over the other? How does editorconfig play into things? Would love to learn more just not sure where to start.

packages/open-truss/src/index.ts
jonmagic:
```suggestion
export * from './utils/format'

```

packages/open-truss/src/utils/format.ts
jonmagic:
```suggestion
}

```

andrew-dillon:
Yep! They work really well together. I used that combo to great effect in pdf-lib:
- [`lint`](https://github.com/Hopding/pdf-lib/blob/93dd36e85aa659a3bca09867d2d8fac172501fbe/package.json#L45C73-L45C73)
- [prettier config](https://github.com/Hopding/pdf-lib/blob/93dd36e85aa659a3bca09867d2d8fac172501fbe/.prettierrc.json)
- [tslint config](https://github.com/Hopding/pdf-lib/blob/93dd36e85aa659a3bca09867d2d8fac172501fbe/tslint.json) (this would be eslint now, tslint is deprecated)

Prettier is more opinionated than editorconfig, so we wouldn't use both.

I'll open a PR adding it to the project. It's pretty straightforward to setup, so if we decide not to merge it then it's no biggy.

PR #33, opened 2023-11-09
Prettier!

andrew-dillon:
Adding Prettier for code formatting as discussed in https://github.com/open-truss/open-truss/pull/27#issuecomment-1802749552.

Most editors have prettier plugins that will automagically format your code on save. It's an incredible experience! You can also run `npm run lint` and Prettier will format all the project's files for you.

In the CI, if files don't match the Prettier/eslint spec, errors/warnings will be printed instead of the files being automagically fixed.

Most of the changes in the PR are diffs from Prettier formatting. The files I added/changed to setup Prettier itself are:
- `packages/open-truss/package.json`
- `packages/open-truss/.prettierrc.json`
- `.github/workflows/ci-checks.yml`
- `packages/open-truss/package-lock.json`

**Note:** If we aren't sure about using Prettier or don't want to merge this PR, that's okay. This didn't take long to setup. Figured it be easier to actually show what using Prettier looks like rather than try to describe it 😄

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -17,7 +17,7 @@ jobs:
         run: npm ci --ignore-scripts
       - name: Run linter
         working-directory: ./packages/open-truss
-        run: npm run lint
+        run: npm run lint:ci
   test:
     runs-on: ubuntu-latest
     steps:
--- a/packages/open-truss/.eslintrc.js
+++ b/packages/open-truss/.eslintrc.js
@@ -1,36 +1,30 @@
 module.exports = {
-    "env": {
-        "browser": true,
-        "es2021": true,
-        "node": true
+  env: {
+    browser: true,
+    es2021: true,
+    node: true,
+  },
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react/recommended',
+  ],
+  overrides: [
+    {
+      env: {
+        node: true,
+      },
+      files: ['.eslintrc.{js,cjs}'],
+      parserOptions: {
+        sourceType: 'script',
+      },
     },
-    "extends": [
-        "eslint:recommended",
-        "plugin:@typescript-eslint/recommended",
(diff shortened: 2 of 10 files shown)


jonmagic (approved):
I LOVE THIS

Question: Can we use prettier at the global project level? I'm asking because I would love to apply [this guidance](https://github.com/open-truss/open-truss/blob/main/docs/pascal-kebab-camel-snake-case.md#naming-files-and-folders) across the entire repository and I'm wondering if a prettier rule could help us achieve that?

andrew-dillon:
We can use it at the project level for sure. I put it in the package since that's where the eslint config was. We could easily move both to the top level. Probably should tbh.

Prettier is strictly a code formatter, so it doesn't know anything about file names. But we could use [this eslint plugin](https://github.com/DukeLuo/eslint-plugin-check-file/blob/main/docs/rules/filename-naming-convention.md) for that:
```js
module.exports = {
  plugins: ['check-file'],
  rules: {
    'check-file/filename-naming-convention': [
      'error',
      {
        '**/*.{jsx,tsx}': 'CAMEL_CASE',
        '**/*.{js,ts}': 'KEBAB_CASE',
      },
    ],
  },
};
```

packages/open-truss/src/index.ts
```diff
@@ -1,3 +1,17 @@
 export { default as helloWorld } from './hello_world'
 export * from './utils/yaml'
 export * from './utils/format'
+class Foo {
```
hktouw, line 4: Are these classes left here intentionally?

hktouw (approved):
Thanks for opening this! A global linter is a great idea.

packages/open-truss/src/index.ts
hktouw: looks like @jonmagic removed those in his PR that branches of this https://github.com/open-truss/open-truss/pull/34/files#diff-f404b0ed66673ec723b8e71bf03bd51f76311501ead6c0f987804eac615dfd75L4.

PR #34, opened 2023-11-10
Tweak linting rules and apply fixes

jonmagic:
Building on and targeting #33 as the base to merge into this PR tries to apply the fixes suggested by the linter but also disables a few lint rules per our preferences.

Definitely open to feedback on whether the rules I disabled make sense to everyone and if there are any others we might want to enable or disable.

- `@typescript-eslint/comma-dangle`: https://typescript-eslint.io/rules/comma-dangle/
- `@typescript-eslint/space-before-function-paren`: https://typescript-eslint.io/rules/space-before-function-paren/
- `react/react-in-jsx-scope`: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md
- `@typescript-eslint/return-await`: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/return-await.md
    - this one is weird, it doesn't error locally but does in CI so I had to disable it 🤷‍♂️

### DIFF
--- a/packages/open-truss/src/utils/format.ts
+++ b/packages/open-truss/src/utils/format.ts
@@ -1,4 +1,4 @@
-import { format as sqlFormat, SqlLanguage } from 'sql-formatter'
+import { format as sqlFormat, type SqlLanguage } from 'sql-formatter'
 
 export type Param = string | number | string[] | number[]
 
@@ -27,12 +27,12 @@ function formatParams(params: Record<string, Param>): Record<string, string> {
   return Object.fromEntries(formattedEntries)
 }
 
-export function formatQuery(template: string, config?: Config) {
+export function formatQuery(template: string, config?: Config): string {
   return sqlFormat(template, {
-    language: config?.language || 'sql',
+    language: config?.language ?? 'sql',
     tabWidth: 4,
     linesBetweenQueries: 2,
     paramTypes: { named: [':'] },
-    params: config?.params && formatParams(config.params),
+    params: config?.params != null ? formatParams(config.params) : undefined,
   })
 }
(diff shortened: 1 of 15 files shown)


packages/open-truss/src/utils/format.ts
```diff
@@ -27,12 +27,12 @@ function formatParams(params: Record<string, Param>): Record<string, string> {
   return Object.fromEntries(formattedEntries)
 }
 
-export function formatQuery(template: string, config?: Config) {
+export function formatQuery(template: string, config?: Config): string {
   return sqlFormat(template, {
-    language: config?.language || 'sql',
+    language: config?.language ?? 'sql',
```
hktouw, line 32: TIL `??`. Really cool

hktouw (approved):
Awesome 🤘

hktouw:
@andrew-dillon looks like the intent is to merge this PR into yours and merge yours into main. I went ahead and reviewed both PRs and things look 👍. At your leisure, feel free to merge the PRs after you've had a chance to take a look.

andrew-dillon:
@kre64 and I turned off a couple additional rules:
- `@typescript-eslint/strict-boolean-expressions` from a super strict standpoint this might make sense, but nobody actually uses `x != null` in their TS/JS code. `!x` is standard and much easier to read. It's never caused me any issues either. Arguably it's safer too, because it handles cases where users pass unexpected values, which is useful since we're building a library.
- `@typescript-eslint/prefer-nullish-coalescing` same idea here. Nullish coalescing is great if you're really confident about what values are being passed in, but since this is a library, we're not. Better to be safe and use the more fault tolerant `||` unless we have a specific reason not to.

PR #35, opened 2023-11-13
Add open-truss bin and tsc_watch command

hktouw:
## Why

We want to expose binaries to package installers that help assist with local development

## How

This PR adds the `open-truss` binary to the open-truss package. When npm packages are installed any binaries / scripts listed in the `bin` portion of the `package.json` will be added. Applications can call the open-truss binary by adding lines such as `"tsc-watch": "npm run open-truss tsc-watch"` to their package.json file.

This PR should be pretty straightforward to review and every file has plenty of comments. You can start with reviewing the `open-truss` bash script and then move on to review the `tsc-watch` command which runs the tsc file watcher on the OT project, recompiling if there are any changes. 

## Next steps

There are more executables to be added to improve the dev workflow, but I thought this was a good place to start. I can open up subsequent PRs adding more. After more are added, I intend on creating docs for all of this.

@open-truss/engineers


packages/open-truss/bin/bin/hello_world
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
andrew-dillon, line 1: Should these files use kebab casing? IIRC that matches the guidelines for this project. It's also much more standard for cli/binary commands in my experience.

andrew-dillon: - `hello_world -> hello-world`
- `tsc_watch -> tsc-watch`

packages/open-truss/bin/open-truss
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
hktouw: When npm packages are installed they add any binaries to the path (global path if installed globally). I felt that name spacing out binaries in  `open-truss` would reduce the likelihood of collisions with other binaries, especially since I'm not sure if any errors are thrown when there are collisions and instead binaries paths may silently get overridden.

hktouw:
@andrew-dillon this is ready for another 👀 when you have a moment

andrew-dillon (approved):

Issue #31, opened 2023-11-09
[Task] Improve our "hybrid" package approach


jonmagic:
I think we should follow most if not all of the advice in https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html. That includes moving the commonjs and esm builds into a `dist` folder instead of having `lib` and `lib-esm` at the root of `packages/open-truss`.

PR #36, opened 2023-11-14
Improve dev env

hktouw:
resolves https://github.com/open-truss/open-truss/issues/31

## Why

@jonmagic and I found a great [blog post](https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html) that gave instructions on how to organize a typescript project like OT in a sane way. We were already following some of the suggestions, but we should implement the rest

## How

This PR implements the rest of the suggestions. Notably, it moves the output javascript files into `dist/cjs` and `dst/mjs` folders and adds `package.json` files to the dist folders. The easiest way to review this PR is commit by commit and reading the comments included, it should be pretty straightforward.

## Testing

I tested these changes in the demo app and application we use at GitHub and everything worked as expected. 

@open-truss/engineers

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -16,12 +16,9 @@ jobs:
       - name: Run CI
         working-directory: ./packages/open-truss
         run: npm ci
-      - name: Build lib
+      - name: Build dist
         working-directory: ./packages/open-truss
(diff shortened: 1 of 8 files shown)


jonmagic (approved):
👍

PR #47, opened 2023-11-29
Introduce a unified query interface

jonmagic:
Introducing the Unified Query Interface. Before this interface handling the results from a Trino, Kusto, MySQL or any data source client required different acrobatics.

The goals of this library include:

1. Handle results from any data source the same way.
2. Support streaming query results out of the box for any data sources that support streaming queries.
3. Make it fairly simple to create a UQI interface for any data source client. NOTE: We plan on creating ones for Trino, Kusto, MySQL, REST, and more in the next few weeks.

To better explain how different each existing data source client is and what this library makes possible let's look at some examples.

## Trino

This example shows setting up a Trino client, executing a query, and processing the results:

```typescript
import { BasicAuth, Iterator, QueryResult, Trino } from 'trino-client'

const trinoClient = Trino.create({
  server: process.env.TRINO_URI,
  auth: new BasicAuth(process.env.TRINO_CREDENTIALS),
  source: 'acme-service',
})

// record start time if stats are needed

try {
  // start the query
  const query: Iterator<QueryResult> = await trinoClient.query('SELECT id, name FROM customers')

  // iterate over the results
  for await (const queryResult of query) {
    // check `queryResult.stats.state` and throw error if FAILED
    
    // retrieve and store column types from `queryResult.columns`
    //   example: [{ name: 'id', type: 'bigint' }, { name: 'name', type: 'varchar' }]

    // a queryResult has a `data` property that is an array of rows, note that this
    // is just one set of rows from the query, there may be more
    if (queryResult.data !== undefined) {
      for (const row of queryResult.data) {
        // store `row` somewhere
        // increment row counter if stats are needed
      }
    }
  }
} catch (error) {
  // update stats with error if stats are needed

  // throw error
  throw error
} finally {
  // record end time if stats are needed
}
```

## Kusto

Here's another example for Kusto:

<details>
  <summary>Click to expand</summary>

```typescript
import { Client, KustoConnectionStringBuilder } from 'azure-kusto-data'

const kcsb = KustoConnectionStringBuilder.withAadApplicationKeyAuthentication(
  process.env.KUSTO_ENDPOINT, // example: https://acme-service.eastus.kusto.windows.net
  process.env.KUSTO_APPID,
  process.env.KUSTO_APP_KEY,
  process.env.KUSTO_AUTHORITY_ID,
)
const client = new Client(kcsb)

// record start time if stats are needed

try {
  const query = await client.execute('acme', 'acme.production.customers | project id, name')

  const status = query.statusTable.toJSON().data.find((row) => row.EventTypeName === 'QueryResourceConsumption')
  // record stats if needed

  for await (const kustoResultRow of query.primaryResults[0].rows()) {
    // retrieve and store column types from `kustoResultRow.columns`
    //   example: [{ name: 'id', type: 'long' }, { name: 'name', type: 'string' }]

    // collect the columns into a row
    const row = []
    for await (const column of kustoResultRow.values()) {
      row.push(column)
    }

    // store `row` somewhere
    // increment row counter if stats are needed
  }
} catch (error) {
  // update stats with error if stats are needed

  // throw error
  throw error
} finally {
  // record end time if stats are needed
}
```
</details>

## MySQL

And another example for MySQL:

<details>
  <summary>Click to expand</summary>

```typescript
import mysql, { FieldPacket, OkPacket, ResultSetHeader, RowDataPacket } from 'mysql2/promise'
import { URL } from 'url'

const client: MysqlPool = ysql.createPool({
  host: process.env.MYSQL_HOST,
  port: Number(process.env.MYSQL_PORT),
  database: process.env.MYSQL_DATABASE,
  connectionLimit: 10, // will make up to this number of concurrent connections to mysql. subsequent requests are queued
  waitForConnections: true,
  user: process.env.MYSQL_USER,
  password: process.env.MYSQL_PASSWORD,
  rowsAsArray: true,
})

// record start time if stats are needed

try {
  const [rows, fields] = await client.execute('SELECT id, name FROM customers')

  // retrieve and store column types from `fields`
  //   example: [{ name: 'id', type: 3 }, { name: 'name', type: 15 }]
  //   - 3 is a LONG and 15 is a VARCHAR, need a lookup table

  // `rows` can be any of the following types:
  //   'RowDataPacket[] | RowDataPacket[][] | OkPacket | OkPacket[] | ResultSetHeader'
  
  if (Array.isArray(rows) && rows.length > 0 && !Array.isArray(rows[0])) {
    // store `rows` somewhere
    // increment row counter if stats are needed
  } else if (Array.isArray(rows) && rows.length > 0 && Array.isArray(rows[0])) {
    for (const row of rows) {
      // store `row` somewhere
      // increment row counter if stats are needed
    }
  } else {
    // store `rows` somewhere
    // increment row counter if stats are needed
  }
} catch (error) {
  // update stats with error if stats are needed

  // throw error
  throw error
} finally {
  // record end time if stats are needed
}
```
</details>

## With UQI

After implementing a uqi client for each of the above data sources, executing a query and processing the results looks the same for all of them:

```typescript
import createClient from 'path/to/<data-source>-uqi-client'

const client = createClient({
  // data source specific options
})

// record start time if stats are needed

try {
  const queryIterator = await client.query('SELECT id, name FROM customers', {
    statusCallback: (status) => {
      // record stats if needed

      // interface UqiStatus {
      //   completedAt: Date | null
      //   failedAt: Date | null
      //   failedReason: string | null
      //   recordsReturned: number
      //   startedAt: Date | null
      // }
    },
  })

  for await (const { row, metadata } of queryIterator) {
    // metadata.columns is an array of normalized javascript types
    //   example: [{ name: 'id', type: 'BigInt' }, { name: 'name', type: 'String' }]

    // store `row` somewhere
  }
} catch (error) {
  // optionally swallow error
}
```

## Creating a uqi client for a data source client

Creating a new uqi client for an existing data source client is only as complicated as the existing clients API:

<details>
  <summary>Click to see an example</summary>

```typescript
import Iterator from '../iterator'
import uqi, {
  type UqiClient,
  type UqiContext,
  type UqiResult,
  type UqiMappedType,
} from '../uqi'
import createFakeClient, {
  type FakeClient,
  type FakeClientConfig,
} from './fake-client'

export default async function (config: FakeClientConfig): Promise<UqiClient> {
  const typeMappings: Record<string, UqiMappedType> = {
    int: 'Number',
    varchar: 'String',
  }

  async function query(
    context: UqiContext<FakeClientConfig, FakeClient>,
    query: string,
  ): Promise<AsyncIterableIterator<UqiResult>> {
    const queryIterator = await context.client.query(query)

    async function* asyncGenerator(): AsyncGenerator<UqiResult> {
      for await (const result of queryIterator) {
        if (result.data) {
          for await (const row of result.data) {
            yield {
              row: [[`Sam ${row[1]}`, row[1]]],
              metadata: result.metadata,
            }
          }
        }
      }
    }

    return new Iterator(asyncGenerator())
  }

  async function teardown(
    context: UqiContext<FakeClientConfig, FakeClient>,
  ): Promise<void> {
    await context.client.close()
  }

  const client = await createFakeClient(config)

  return uqi({
    typeMappings,
    config,
    client,
    query,
    teardown,
  })
}
```
</details>

## Reviewing this PR

I would not review commit by commit as there was quite a bit of back and forth before landing on the a design that felt right.

Instead review files in this order:
- [The README](https://github.com/open-truss/open-truss/pull/47/files#diff-4e74546b6b004d2ec7d4cfe549fc3e462930dcdbe248161e6afeabbe588e3063)
- [A fake-client implementation](https://github.com/open-truss/open-truss/pull/47/files#diff-bb0061c2b53c1166e52e1fa495a965b8338b0caad945a17640aa9f12b26987d1) and [it's tests](https://github.com/open-truss/open-truss/pull/47/files#diff-81c8cc79f4655fa4506308b83c9f58c4004a6e3489bb20f1713c03d238b862d0) that was modeled after [trino-client](https://github.com/regadas/trino-js-client) (mostly).
- After looking at the fake-client check out [uqi](https://github.com/open-truss/open-truss/pull/47/files#diff-6a87aa09e21fc4975a92032bf0fd706f633791edc5a6d265d02263f55ea57643).
- Now look at [the example of wrapping the fake-client with the uqi-client](https://github.com/open-truss/open-truss/pull/47/files#diff-40ba48a9fd99b0204b4c50a10dd6b3b1a2540be7f9d732fa39f307056a17bfd2) and [the test](https://github.com/open-truss/open-truss/pull/47/files#diff-dee9c656afa692d7fa4b1d6932f1f41af5618ea9f097dc5e82de7b72e3d15c9d) showing it in action.


packages/uqi/README.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+### Querying a Rest API
+
+```typescript
+import { uqi } from '@open-truss/uqi'
+const source = uqi.rest({
+  url: 'http://localhost:3000',
+})
+for await (const { row, metadata } of source.query('GET /users')) {
```
kmcq, line 86: REST and JSON are interesting cases because they might not be nicely enumerable like the rest -- a JSON object could be an object and not an array, a REST endpoint might return a single entity.

Do you think there's another single-entity API? Or, would a `for` loop not be used? Or, would uqi wrap non-array results to be an array?

I think my favorite is to just not use a loop.

kmcq (approved):
Great name, great examples!

packages/open-truss/src/uqi/README.md
```diff
@@ -0,0 +1,116 @@
...
+It is a query interface with pluggable adapters that makes it simple to query different data sources from a single interface.
+
+## Implemented
+
+### Type transformations
+
+OpenTruss will support a number of query engine backends and data sources. Each of these will have their own type system. OpenTruss will provide a way to transform these types to javascript types so that workflows can be built on top of the simpler javascript types.
+
+```typescript
+otType('mysql', 'int')
```
hktouw, line 16: `otDataType`? `otScalar`? `otScalarType`?

hktouw (approved):
Really like how this is coming along 🤘. Do you think this is good to ship as is and we can iterate on it on a follow up PR?

packages/open-truss/src/uqi/README.md
jonmagic: otDataType is nice 👍

jonmagic:
> Really like how this is coming along 🤘. Do you think this is good to ship as is and we can iterate on it on a follow up PR?

Imma keep working on it in this PR. I'm spiking on the query interface now.

jonmagic:
As I look at https://github.com/open-truss/open-truss/blob/6721443318ea6b37bf85532b5adcdbcaaf34ee99/packages/open-truss/src/uqi/example/fake-client-uqi.test.ts#L9-L14 I'm realizing I would prefer to pass the config to the createFooUqiClient function instead of calling client.setup with the config later 🤔

I'll make this API change and then request review again, for now I"m going to remove the review request.

packages/open-truss/src/uqi/iterator.ts
```diff
@@ -0,0 +1,64 @@
+export type IteratorResult<T> = Promise<{
```
hktouw, line 1: I know this is probably just implementing the iterator interface, but this file is 🤯. Is there a reason you needed to do this? Are there no standard library utilities we could use / inherit from?

packages/open-truss/src/uqi/uqi.ts
```diff
@@ -0,0 +1,176 @@
...
+      context.status.startedAt = new Date()
+      if (options?.statusCallback) {
+        await options.statusCallback(context.status)
+      }
+
+      async function* asyncGenerator(): AsyncGenerator<UqiResult> {
+        try {
+          for await (const result of queryIterator) {
+            const row = buildRow(result)
+            const metadata = buildMetadata(result.metadata)
```
hktouw, line 143: I assume the answer is yes, but will there be cases where each row has different metadata? Kind of unfortunate to have to recalculate this every time.

packages/open-truss/src/uqi/example/fake-client-uqi.ts
```diff
@@ -0,0 +1,56 @@
...
+  type UqiResult,
+  type UqiMappedType,
+} from '../uqi'
+import createFakeClient, {
+  type FakeClient,
+  type FakeClientConfig,
+} from './fake-client'
+
+export default async function (config: FakeClientConfig): Promise<UqiClient> {
+  const typeMappings: Record<string, UqiMappedType> = {
```
hktouw, line 14: That's a little annoying you have to declare the type here instead of having it be implicit. I feel like typescript should be smarter about knowing when an untyped object adheres to the interface of the function you are passing it to.

packages/open-truss/src/uqi/example/README.md
```diff
@@ -0,0 +1,7 @@
+# Unified Query Interface Example
+
+First check out the fake-client implementation and tests.
+
+Then check out the fake-client-uqi implementation and tests.
+
+We'll be providing more example implementations for Trino, Kusto, MySQL, REST, and more in the future.
```
hktouw, line 7: 👍 

Are you thinking that OT will have its own implementations of all those clients? I know you want to avoid the maintenance overhead of that, but it could be nice to have that open sourced. Perhaps we can set expectations around support for keeping those clients updated and recommending people implement their own or consider contributing to OT?

packages/open-truss/src/uqi/uqi.ts
```diff
@@ -0,0 +1,176 @@
+import Iterator from './iterator'
+
+export type UqiMappedType =
```
hktouw, line 3: I forget, we decided to not support ArrayBuffers correct? Maybe worth adding a brief comment here saying that is intentional?

That being said, I feel like we'll have to support that sooner than later to process files. But that seems like a problem to tackle when we get there.

packages/open-truss/src/uqi/uqi.ts
```diff
@@ -0,0 +1,176 @@
...
+    return result.metadata.columns.map((column: UqiColumn, i: number) => {
+      const type = context.typeMappings[column.type]
+      if (!type) {
+        throw new Error(`Type ${column.type} is not mapped`)
+      }
+      const value = result.row[i]
+      if (value === null) {
+        return null
+      }
+      if (typeof value === 'string' && type === 'Number') {
```
hktouw, line 85: How do you think this logic will work when we have untyped data sources like rest apis?

hktouw (approved):
Love this 🤘. I appreciate the thoughtfulness of this design and well-made architecture.

I left a few comments, but I think this is ready to ship. This seems like something we can ship, see how it works, then iterate on.

packages/open-truss/src/uqi/iterator.ts
jonmagic: > Is there a reason you needed to do this?

I actually stole this from the `trino-client` library (I'll add attribution to the top of the file). It is just a wrapper around  `AsyncIterableIterator` that can be extended to include metadata like the id of a query.

> Are there no standard library utilities we could use / inherit from?

Possible?
- https://github.com/async-generators/iterable was archived in 2019 and doesn't allow extension
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/Iterator looks very similar but is experimental

packages/open-truss/src/uqi/uqi.ts
jonmagic: Fwiw I wasn't going to start optimizing performance on any parts of this (besides using the iterator/generator pattern) until we needed to. That said I think only calculating metadata when needed makes sense.

I think we have two options:
1. Update the API to return a function instead of the already calculated metadata so that you can optionally build the metadata when you need it.
2. Cache or memoize the calculated metadata.

Preference?

packages/open-truss/src/uqi/example/fake-client-uqi.ts
jonmagic: Maybe I did it un-necessarily? Not sure, maybe you could hack on it a bit to see if there is a better way?

packages/open-truss/src/uqi/example/README.md
jonmagic: I'm thinking the demo-app will have implementations of all of these clients to start and then we can pull them out into packages. We need to figure out our story around publishing many packages first though.

packages/open-truss/src/uqi/uqi.ts
jonmagic: I'd like to support them sooner rather than later but this PR was already a beast so thought it could be an addition in the near future once I got to a data source that required it.

packages/open-truss/src/uqi/uqi.ts
jonmagic: I don't know but I'm excited to find out! I'm guessing when someone is using a REST API they'll have to define an interface of some sort ahead of time that gets passed in the create function for the uqi rest client. Or maybe we have to bring in zod 🤔

packages/open-truss/src/uqi/uqi.ts
hktouw: number 2 makes sense to me. can probably cache it based on the metadata object. but also 👍 for optimizing later.

packages/open-truss/src/uqi/example/fake-client-uqi.ts
hktouw: I tried removing the type locally and it wasn't happy :/

packages/open-truss/src/uqi/uqi.ts
kmcq: Specifying the type of a rest endpoint is what I think 👍

Issue #32, opened 2023-11-09
[Task] Build a component/workflow/rendering playground


jonmagic:
This playground will make it easier and faster for us to iterate on the:

1. Workflow spec
2. Components
3. Rendering engine

hktouw:
We shipped a simple, yet function version of a playground. https://github.com/open-truss/open-truss/pull/48. This should serve as a good base as we build out the rendering engine and iterate on components. 

@kmcq @jonmagic There is some work to be done to ship a fully functioning version of the rendering engine with zod types. Do you think that should be in a follow up issue or a part of this one?

kmcq:
I'm currently working faster and more chaotic than an issue, but I'll be slowing down soon and can write up what is in my head that I don't get to so that it's somewhere more durable

kmcq:
Okay it's time for me to slow down a bit. I think that the following recent PRs wrap up this task:

- https://github.com/open-truss/open-truss/pull/64
- #76 
- #77

We now have a zod-powered way to construct and view configs in real time! I'll write up another issue for what else we could do in this space.

PR #48, opened 2023-11-29
Rendering engine V1 POC

hktouw:
ref https://github.com/open-truss/open-truss/issues/32

## Why

We want a rendering engine so that we can take a workflow config and render components

## How

This PR is a first pass at setting up the project to support a rendering engine. This:
- Sets up the project to make versioning the rendering engine easy. This will be necessary as we will likely need to do multiple iterations to get this right.
- Implements a POC rendering engine. To get farther we will need zod + runtime types.
- Makes component directories and auto imports all components so that they are available to the rendering engine.
- Makes a playground page to execute a sample yaml config.

The best way to review this PR is by going line by line in starting with the demo-app playground page. If you step through the execution flow it should be clear how the system is working. It may be easiest to use your editor to jump to definitions. Don't forget to compile the typedefs using `npm run tsc:watch` in the demo app project.

## Testing

Rendered the playground page and it seemed to work just fine.

Model Home 2023-11-29 at 3 33 10 PM

## Follow ups

- *Rendering engine V2* - Using @kmcq's great work on zod runtime types, we can iterate on the V2 engine that will allow us to add type checking to our data or yaml configs dynamically. The V2 engine can also explore other ideas such as using jotai / signals as a state store to share data between components.
- *Better playground features* - In a follow up PR I'll add more features to the playground to make it easier to iterate on display components. e.g. adding a mock data file that gets imported into the rendering engine, multiple playground pages to make it easier to try different things and save our work for future review, etc. @kmcq It will be interesting to think about what a future UX might look like that leverages both the playground and UI builder you're working on.

cc @open-truss/engineers


packages/open-truss/src/components/index.ts
```diff
@@ -0,0 +1 @@
+export { default as OTBar } from './OTBar'
```
hktouw, line 1: To avoid name collisions with components developed in applications, I prefixed OT components with OT. What do folks think of this?

packages/open-truss/src/configuration/apply.tsx
```diff
@@ -0,0 +1,34 @@
...
+export type RenderingEngine = () => ReactTree
+
+type ConfigurationFunction = (config: YamlObject, data: YamlType) => ReactTree
+export function applyConfiguration(COMPONENTS: COMPONENTS): ConfigurationFunction {
+  const components = Object.assign(COMPONENTS, OTCOMPONENTS)
+
+  const configurationFunction: ConfigurationFunction = (config, data) => {
+    let renderingEngine
+
+    const workflow = (config as unknown as WorkflowSpec).workflow
```
hktouw, line 21: This might benefit from a zod runtime type check

packages/open-truss/src/utils/yaml.ts
```diff
@@ -1,9 +1,15 @@
 import yaml from 'yaml'
 
-interface YamlObject extends Record<string, YamlType> {}
-type YamlType = null | number | string | boolean | YamlType[] | YamlObject
+export interface YamlObject extends Record<string, YamlType> {}
+export type YamlType =
```
hktouw, line 4: The linter yelled at me to unfurl this

jonmagic:
To get it running I had to:

1. `cd packages/open-truss && npm install`
2. `cd demo-app && npm install`
3. `cd demo-app && npm run tsc:watch`
4. Open browser to http://localhost:3000/playground

packages/open-truss/src/components/index.ts
jonmagic: Great idea

jonmagic (approved):
:sparkles:

PR #51, opened 2023-12-01
[WIP] Open Truss application ADR

hktouw:
🚧 WIP 🚧

This is a work in progress, but folks are welcome to leave comments and ask questions. 

/cc @open-truss/engineers

PR #55, opened 2023-12-14
Setup nextjs OT application

hktouw:
## Why

As described in the [ADR](https://github.com/open-truss/open-truss/pull/51), we want to develop server side features for Open Truss. This requires us to update the structure of this project to be fullstack.

## How

I initially [spiked out a middleware](https://github.com/open-truss/open-truss/pull/53) implementation of OT, but that ran into issues. I then made a [spike where ot is a full nextjs app](https://github.com/open-truss/open-truss/pull/54), but that also ran into issues.

This PR tries an approach where OT has a full nextjs app, but we silo OT specific functionality into specific directories. e.g. `src/pages/ot`. Then we have a script that copies all the ot directories into user applications. I imagine there will be some massaging that needs to happen after instal to get this to work for apps, but I believe applications can be refactored to make this simple. Additionally, we can set this up to where it is mostly automated for apps that follow standard nextjs (or frameworks) conventions.

The easiest way to review this PR is commit by commit starting with this https://github.com/open-truss/open-truss/commit/00df75b4870b238e372c063161fd69f097582b0c and reading the comments along the way. The first commit is a development Readme and should provide a good overview of how the project is setup.

## Next steps

- @open-truss/engineers your feedback is requested.
- If this approach seems reasonable, I would like to merge this PR and continue spiking this out in a follow up PR. I want to try to get this setup in GitHub and see what other hurdles I run into.


docs/development.md
```diff
@@ -0,0 +1,74 @@
...
+### Developing Open Truss libraries for your application
+
+Same as above
+
+### Developing the Open Truss application
+
+Setup:
+
+1. Complete app instillation above
+1. Complete open truss component instillation above
```
jonmagic, line 55:
```suggestion
1. Complete app installation above
1. Complete open truss component installation above
```

docs/development.md
```diff
@@ -0,0 +1,74 @@
+## Development
+
+## Application Installation
+
+Open Truss is a full-stack framework installable into existing applications. Open Truss comes with scripts to generate client and server files, adding them to your application.
+
+OT supports the following applications:
```
jonmagic, line 7:
```suggestion
Open Truss supports installation into applications built with the following frameworks:
```

docs/development.md
```diff
@@ -0,0 +1,74 @@
...
+- NextJS
+
+### NextJS
+
+This assumes a standard typescript NextJS app structure with a `src` directory.
+
+1. Run `npm install @open-truss/open-truss`
+1. Copy [open-truss.conf.default.yaml](../packages/open-truss/nextjs/open-truss.conf.default.yaml) into NextJS root and name `open-truss.conf.yaml`
+1. Add `"ot:setup": "open-truss setup"` to package.json and run
+  **Important** - this adds files to your NextJS application, overwriting them if they exist. These are namespaced (e.g. `ot`) and should not conflict with your application-specific files. See [bin/setup](../packages/open-truss/bin/setup) for the full list of files added.
```
jonmagic, line 17: The wording here is a tad confusing. Maybe mention namespacing before the overwriting warning since overwriting is less likely due to namespacing?

docs/development.md
```diff
@@ -0,0 +1,74 @@
...
+1. Run `npm install`
+1. Run `script/server`
+1. Start building nextjs application pages
+1. Run `open-truss setup` in your application when you want the files copied over.
+
+`open-truss setup` copies files from open-truss's application (e.g. nextjs) into folders inside of the user's application. If you are modifying files that are not part of directories that are already being copied, you can add them here. [bin/setup](../packages/open-truss/bin/setup)
+
+### Notes & FAQs
+
+- **tsc:watch gotchas** - `tsc:watch` watches for source file changes and automatically generates js from typescript. When you rename or delete source files, it _does not_ delete the old compiled js files. This is usually fine, but sometimes can cause bugs. When in doubt, you can delete the `dist` directory and restart `tsc:watch`.
```
jonmagic, line 66:
```suggestion
- **tsc:watch gotchas** - `tsc:watch` watches for source file changes and automatically generates javascript from typescript. When you rename or delete source files, it _does not_ delete the old compiled javascript files. This is usually fine, but sometimes can cause bugs. When in doubt, you can delete the `dist` directory and restart `tsc:watch`.
```

packages/open-truss/bin/setup
jonmagic, line 1: Can we use 2 spaces indent instead of 4 spaces?

packages/open-truss/bin/setup
```diff
@@ -0,0 +1,64 @@
...
+}
+
+
+function main() {
+    const openTrussConf = getOpenTrussConf()
+
+    copyOTFilesToProject(openTrussConf)
+}
+
+main()
```
jonmagic, line 64:
```suggestion
main()

```

packages/open-truss/nextjs/README.md
jonmagic, line 1: Do we need this file?

jonmagic (commented):
Left some notes inline but wanted to start a conversation here as well. Please take my feedback with a grain of salt, I'm recovering from covid and probably not thinking clearly.

I'm unclear what this PR solves right now, for me at least. For current development I need to be able to iterate on `packages/open-truss` and use the `demo-app` as the place to show off the interplay between the open-truss package and an application.

I'm concerned that scripting away the installation of the integration of the open-truss package and the applications that use it will be very difficult to maintain. For example, every time a new version of create-next-app is released we'll have to update `packages/open-truss/nextjs` and if we support installation into multiple frameworks we'll have to manually update all of those as well.

My thinking was that the `demo-app` could be an example that users look to for inspiration on how to install and integrate our packages into their applications. We can link from our documentation to specific lines in the demo-app, which they can fire up locally to see it in action, modify, and so on.

So I guess I'm wondering if we can update this PR to only make the changes to the `demo-app` that introduce the `/ot/` routes and not to `packages/open-truss` OR to move the changes to `packages/open-truss` over to a new package like `packages/install-ot-integration`?

I apologize if I missed something in my haze 🤒

hktouw:
@jonmagic great questions!

> I'm unclear what this PR solves right now
> My thinking was that the demo-app could be an example that users look to for inspiration on how to install and integrate our packages into their applications.

This solves the problem where we need to build server side features for OT (e.g. config service, query engine, workflows, etc). We should make those all individual libraries in OT, but it would be even easier if OT also had control over entire APIs and pages to make the integration smoother.

We would have to build those anyway for our app, so might as well do that in an open source way. e.g. for your uqi work, you could build a uqi Rest / Graphql API in the OT nextjs app and that can be copied over to the nextjs app we have. That's more reusable than implementing it in our app.

I view this as similar to how Rails approaches things. You install the gem and generate some code. Ours does the same. The main difference is our generated code is outputted in "user space" whereas rails is in the vendor code, but it's still the same concept.

> I'm concerned that scripting away the installation of the integration of the open-truss package and the applications that use it will be very difficult to maintain. For example, every time a new version of create-next-app is released we'll have to update packages/open-truss/nextjs and if we support installation into multiple frameworks we'll have to manually update all of those as well.

I think that is something to keep track of, but I feel pretty confident we can keep our footprint relatively small. Much of this code can be libraries in OT and the platform specific bits (e.g. nextjs) are just wrappers. Every page for example is an OT page so most the view code should be in the components library and not in the app. And if we make the server side configurable too, then the same applies. And I'm not too worried about maintaining multiple frameworks. We'll maintain the one we care about and if others want to maintain a separate one they are more than welcome to.

> I apologize if I missed something in my haze 🤒

:heart:

docs/development.md
hktouw:
```suggestion
  **Important** - this adds open-truss namespaced files to your NextJS application, overwriting them if they exist. These are in prefixed with `open-truss` or in directories such as `ot` and should not conflict with your application-specific files. See [bin/setup](../packages/open-truss/bin/setup) for the full list of files added.
```

packages/open-truss/bin/setup
hktouw: Sounds good! This was mostly generated by copilot so it has a bias towards 4.

packages/open-truss/nextjs/README.md
hktouw: oh sure this can be deleted. this is just what came with the nextjs install

packages/open-truss/bin/setup
jonmagic: fixed in efb9cf4a0806d386d1605d934905f26553741653

jonmagic (approved):
I'm still not completely sold on the approach but I went ahead and applied my last requested changes and am approving so I can merge 👍

hktouw:
> I'm still not completely sold on the approach but I went ahead and applied my last requested changes and am approving so I can merge

Totally fine 👍. Let's keep playing around with it and if nothing else we'll learn a lot. Anything we build that we want to bring over can be easily copied. FWIW though, I'm feeling pretty optimistic about this approach but we'll see how it goes. :)

PR #56, opened 2023-12-17
Improve uqi while adding trinoUqiClient

jonmagic:
I started this PR by spiking on a trinoUqiClient in the demo-app. 

As I built out the trinoUqiClient I had to make some updates to uqi in the `@open-truss/open-truss` package.

One large change I made that was talked about in the previous uqi PR [here]() was to have uqi results be returned with named fields. So instead of getting back a row like this with the field names omitted:

```
[['Jon', 42], ['Jane', 52]]
```

By default you get back this:

```
[{ name: 'Jon', age: 42 }, { name: 'Jane', age: 52 }]
```

You can pass an option (`{ namedFields: false }`) to get back the original representation.

To manually test drive things I updated the `/ot/playground` route temporarily to execute a trino query (I installed and started trino using [these instructions](https://trino.io/docs/current/installation/containers.html#running-the-container)) and render the results.

Model Home 2023-12-17 at 13 59 58


demo-app/src/app/ot/playground/page.tsx
```diff
@@ -1,12 +1,50 @@
...
+  const queryIterator = await trinoUqiClient.query(
+    'select * from  tpch.sf1.customer limit 10',
+  )
+  const rows = []
+  let metadata: UqiMetadata = {
+    columns: [],
+  }
+  for await (const { row, metadata: m } of queryIterator) {
+    rows.push(row)
+    if (metadata.columns.length === 0) {
```
hktouw, line 25: Out of curiosity what is this if block for? Just for demo purposes?

demo-app/src/app/ot/playground/page.tsx
```diff
@@ -1,12 +1,50 @@
...
+import createTrinoUqiClient from '@/lib/trino-uqi-client'
+
 const configurationFunction = applyConfiguration(COMPONENTS)
 
 async function PlaygroundPage(): Promise<JSX.Element> {
+  const trinoUqiClient = await createTrinoUqiClient({
+    auth: 'trino',
+    server: 'http://trino.orb.local:8080',
+  })
+  const queryIterator = await trinoUqiClient.query(
```
hktouw, line 16: Love this. Really slick / clean API and it's great that is going to be consistent across backend types

demo-app/src/app/ot/playground/page.tsx
```diff
@@ -1,12 +1,50 @@
-import { parseYaml, applyConfiguration } from '@open-truss/open-truss'
```
hktouw, line 1: After my [playground updates PR](https://github.com/open-truss/open-truss/pull/57) merges, what do you think about making this into a demo-app OT component and have it rendered with a new config? 

e.g. you can make an `demo-app/src/open-truss/components/trino-client-demo.tsx` component and config called `demo-app/src/open-truss/configs/2-trino-client-demo.yaml` to the effect of.

```yaml
workflow:
  version: 1
  frames:
    - frame:
      view:
        component: TrinoClientDemo
```

hktouw (approved):
This looks fantastic! Phenomenal work 🤘

demo-app/src/app/ot/playground/page.tsx
jonmagic: I'm still trying to solve the problem where the client might want to keep track of metadata like column types but it only needs to capture that data once. This was a hack to do that.

PR #57, opened 2023-12-19
Update playground

hktouw:
## Why

We want to make it easy to iterate on configs using the playground

## How

This PR updates the playground to support multiple configs. Now you can add yaml configs to the `src/open-truss/configs` directory and they will be rendered at `/ot/playground/your-config-name`. You can test this by running `script/server` in the demo-app.

### `/ot/playground`

Model Home 2023-12-19 at 2 56 16 PM

### `/ot/playground/1-foo-bar`

Model Home 2023-12-19 at 2 57 25 PM

Also, this PR modifies the open-truss app to have a main `src/open-truss` folder for the nextjs app. That will give us and the user a central point to put all open-truss related code. Currently I added only the `open-truss/configs` and `open-truss/components` directories, both of which are intended to be modified by the user. Eventually I think we can add an `open-truss/vendor` directory where we can store open-truss related code that is not intended to be touched by users and we can modify freely. 

The easiest way to review this PR is commit by commit, reading the comments along the way. It should be pretty straightforward.

## Considerations

@open-truss/engineers what do you all think about the following conventions?
- Prefixing our config names with an ascending number? e.g. `src/open-truss/configs/1-foo-config.yaml`. That way there is a bit of order to them like a database id.
- Being very liberal with iterating on configs and committing them to the demo-app source (and at GitHub source once we add the playground there). I don't think there is a downside to this at it makes it easy for us to start collecting a lot of different examples. I think this will be especially nice as we build out the rendering engine and it is going to change a lot. It will be useful for folks to have reference points to look back to.


demo-app/src/app/ot/playground/page.tsx
```diff
@@ -1,17 +1,14 @@
...
-  return <>{renderedComponents}</>
+  return <div style={{ whiteSpace: 'pre' }}>{config}</div>
 }
 
 export default PlaygroundPage
 
 const config = `
+To use the playground you can create a new Open Truss configuration at src/open-truss/configs/1-your-config.yaml. For example:
+
+src/open-truss/configs/1-foo-bar.yaml
```
hktouw, line 10: Hmmm. Perhaps we should put playground configs in their own directory as distinct from normal configs?

Maybe `src/open-truss/configs/playground/1-foo-bar.yaml`?

packages/open-truss/nextjs/src/app/ot/playground/[slug]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+  let config
+  try {
+    const sanitizedSlug = path.basename(slug)
+    config = await fs.readFile(
+      `./src/open-truss/configs/${sanitizedSlug}.yaml`,
+      'utf-8',
+    )
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
+      return <>404 - File Not Found</>
```
Comment (jonmagic) [inline packages/open-truss/nextjs/src/app/ot/playground/[slug]/page.tsx:26]: If we're going to 404 here we should probably legitimately 404. I'll add a commit that uses these instructions https://nextjs.org/docs/pages/building-your-application/routing/custom-error#reusing-the-built-in-error-page

packages/open-truss/nextjs/src/app/ot/playground/[slug]/page.tsx
```diff
@@ -0,0 +1,36 @@
+import { promises as fs } from 'fs'
+import { parseYaml, applyConfiguration } from '@open-truss/open-truss'
+import * as COMPONENTS from '@/open-truss/components'
+import path from 'path'
```
Comment (jonmagic) [inline packages/open-truss/nextjs/src/app/ot/playground/[slug]/page.tsx:4]: These aren't sorted correctly.

demo-app/src/app/ot/playground/[slug]/page.tsx
```diff
@@ -0,0 +1,36 @@
+import { promises as fs } from 'fs'
+import { parseYaml, applyConfiguration } from '@open-truss/open-truss'
+import * as COMPONENTS from '@/open-truss/components'
+import path from 'path'
```
Comment (jonmagic) [inline demo-app/src/app/ot/playground/[slug]/page.tsx:4]: These aren't sorted correctly.

demo-app/src/app/ot/playground/[slug]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+  let config
+  try {
+    const sanitizedSlug = path.basename(slug)
+    config = await fs.readFile(
+      `./src/open-truss/configs/${sanitizedSlug}.yaml`,
+      'utf-8',
+    )
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
+      return <>404 - File Not Found</>
```
Comment (jonmagic) [inline demo-app/src/app/ot/playground/[slug]/page.tsx:26]: If we're going to 404 here we should probably legitimately 404. I'll add a commit that uses these instructions https://nextjs.org/docs/pages/building-your-application/routing/custom-error#reusing-the-built-in-error-page

jonmagic (approved):
I like these changes 👍 I do have some notes and I'll be applying some changes before merging. Also this:

It's obvious to me now why I was uncomfortable with https://github.com/open-truss/open-truss/pull/55. I think it was a pre-mature optimization to try and build the scaffolding in packages/open-truss.

In this PR you had to change things in multiple places (scaffolding and demo-app) which meant I hard to review them in multiple places and apply changes in multiple places. At some point I recognize we'll have to do processes like these but right now we should be moving quickly.

I suggest we remove the scaffolding for now and add it later when we have folks wanting to use the library other than ourselves.

demo-app/src/app/ot/playground/[slug]/page.tsx
Comment (jonmagic) [reply demo-app/src/app/ot/playground/[slug]/page.tsx]: Fixed in dc2feea99b4c22892ba8733f664d75c19b44994c

packages/open-truss/nextjs/src/app/ot/playground/[slug]/page.tsx
Comment (jonmagic) [reply packages/open-truss/nextjs/src/app/ot/playground/[slug]/page.tsx]: Those ended up being instructions for client side errors. I had to dig a bit to find the actual answer for rsc: https://github.com/vercel/next.js/discussions/53225

Fixed in 90280cff5e6aedc95a27dea6a9e0714b19acdedd

kmcq:
> change things in multiple places

@jonmagic I agree with this, but I think I'm actually okay with it if we say that `demo-app/src/app/ot` should _only_ be changed by running `npm run ot:setup`. We could even gitignore it if you'd like and change `npm run dev` to first run that command.

Alternatively, we could do what you're suggesting, which I think is undo parts of #55 and build the demo app without the code being in the OT package until we start wanting the package to give us these things. I think I'm okay with either. I have a PR coming soon that updates the demo app's README where we can discuss this as I talk about this in it.

demo-app/src/app/ot/playground/[slug]/page.tsx
```diff
@@ -0,0 +1,37 @@
+import * as COMPONENTS from '@/open-truss/components'
+import { applyConfiguration, parseYaml } from '@open-truss/open-truss'
+import { promises as fs } from 'fs'
+import { notFound } from 'next/navigation'
```
Comment (kmcq) [inline demo-app/src/app/ot/playground/[slug]/page.tsx:4]: This is a very interesting addition. I had wanted to move this file essentially into the OT package, but this would mean that OT depends on `next` and I don't think we want that 🤔

hktouw:
> change things in multiple places

I agree that that makes the dev flow a little tedious. I think the main advantage though is facilitating reuse between Keith and Us. If I built the playground page at GitHub, Keith would have to rebuild it again. This is going to keep happening for every page and API we build that doesn't fit neatly into a package. And the more drift there is between our apps the harder it is to share things we build and the more coordination we have to do.

Another idea to make local dev easier is to move most of our development away from the demo-app and into the `packages/open-truss/nextjs` project. There is a fully bootable nextjs app there and we can easily set it up with the same scripts and workflows we have in the `demo-app`. We would then only use the `demo-app` for testing the final integration into apps and for validating the `ot:setup` script.

PR #59, opened 2023-12-20
Demo application development is done through the OT package

kmcq:
## Why?

I wanted to make the demo app in line with current standards using `npm run ot:setup`.

## How?

- Gave example components more descriptive names and cleaned up example config
- Refactor RenderFromFile to be inside the OT package
  - This was not really related, sorry! Was just something else I was working on.
- Update demo app README to be about Model Home and how we develop it
- Remove legacy pieces of demo app and run ot:setup

PR #60, opened 2023-12-20
Update rendering engine to support async components

hktouw:
## Why

To leverage React Server Components we need the ability to support async open truss components

## How

This PR updates the type defs of our rendering engine so that it supports async open truss components. Some things of note:
- We ran into issues where calling `<Component key={i} data={data} config={config} {...props} />` would throw type errors because we updated the type to where Component could now also be a `Promise<JSX.Element>` in addition to `JSX.Element`. It took a while but I figured out the issue was `<Component />` was the incorrect way to call a promise. Instead this code just calls `Component` as a function instead of the syntactic sugar and `awaits` the promise.
- I dropped the usage of `React.FC<>`. After googling around I found that this is no longer the recommended way to declare React component types. The particular issue we were running into was the `React.FC<>` type does not support async function components. This is because it was made at a time when components needed to be synchronous.

@kmcq This PR updates the interface for OT components. We may want to merge this PR first then update [your PR](https://github.com/open-truss/open-truss/pull/59) to use the new interface. @jonmagic this may apply to your trino demo PR as well.

### DIFF
--- a/packages/open-truss/src/configuration/engine-v1.tsx
+++ b/packages/open-truss/src/configuration/engine-v1.tsx
@@ -31,24 +31,29 @@ export function engineV1(
   config: WorkflowV1,
 ): RenderingEngine {
   const renderFrames = async (frames: FrameV1[]): Promise<ReactTree> => {
-    return (() => {
-      return frames.map(async ({ view, data, frames: subFrame }, i) => {
-        const { component, props: viewProps } = view
-        const Component = COMPONENTS[component]
-        const props = {
-          key: i,
-          data,
-          config,
-          ...viewProps,
-        }
+    return (async () => {
+      return Promise.all(
+        frames.map(async ({ view, data, frames: subFrame }, i) => {
+          const { component, props: viewProps } = view
+          const Component = COMPONENTS[component]
+          const props = {
+            key: i,
+            data,
+            config,
+            ...viewProps,
+          }
 
-        if (subFrame === undefined) {
-          return await Component({ ...props })
-        } else {
-          const children = <>{renderFrames(subFrame)}</>
-          return await Component({ ...props, children })
-        }
-      })
+          if (subFrame === undefined) {
+            return await Component({ ...props })
+          } else {
+            const subFrames = (await renderFrames(subFrame)).map((child, k) => {
+              return <div key={k}>{child}</div>
+            })
+            const children = <>{subFrames}</>
+            return await Component({ ...props, children })
+          }
+        }),
+      )
     })()
   }
 
--- a/packages/open-truss/src/configuration/engine-v1.tsx
+++ b/packages/open-truss/src/configuration/engine-v1.tsx
@@ -29,7 +29,6 @@ export interface WorkflowV1 {
 export function engineV1(
   COMPONENTS: COMPONENTS,
   config: WorkflowV1,
-  data: YamlType,
 ): RenderingEngine {
   const renderFrames = async (frames: FrameV1[]): Promise<ReactTree> => {
     return (() => {
--- a/packages/open-truss/src/configuration/engine-v1.tsx
+++ b/packages/open-truss/src/configuration/engine-v1.tsx
@@ -1,12 +1,16 @@
 import type { YamlObject, YamlType } from '@/utils/yaml'
-import React from 'react'
 import { type RenderingEngine, type ReactTree, type COMPONENTS } from './apply'
 
-export interface BaseOpenTrussComponentV1 {
-  children?: React.ReactNode
+export interface BaseOpenTrussComponentV1Props {
+  key?: number
+  children?: JSX.Element | Promise<JSX.Element>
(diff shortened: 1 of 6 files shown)


packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -27,21 +31,29 @@ export function engineV1(
...
-        <Component key={i} data={data} config={config} {...props} />
-      ) : (
-        <Component key={i} data={data} config={config} {...props}>
-          {renderFrames(subFrame)}
-        </Component>
-      )
-    })
+  const renderFrames = async (frames: FrameV1[]): Promise<ReactTree> => {
+    return (() => {
+      return frames.map(async ({ view, data, frames: subFrame }, i) => {
```
jonmagic, line 36: I think you’ll want to wrap this in a Promise.all maybe? Typically keeps things in order when you map over an array of promises.

jonmagic (approved):
Nice work!

kmcq:
On my phone so I can't check types but it looks like there are some functions in here that are declared with 'async function' but don't use 'await' in them. I thought that didn't do anything, does it actually?

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -27,21 +31,29 @@ export function engineV1(
   config: WorkflowV1,
   data: YamlType,
```
kmcq, line 32: Not new to this PR but it looks like this `data` isn't used and the name is overridden on line 31 in `(config, data) => {` so perhaps we should remove it up here?

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -27,21 +31,29 @@ export function engineV1(
...
-      return subFrame === undefined ? (
-        <Component key={i} data={data} config={config} {...props} />
-      ) : (
-        <Component key={i} data={data} config={config} {...props}>
-          {renderFrames(subFrame)}
-        </Component>
-      )
-    })
+  const renderFrames = async (frames: FrameV1[]): Promise<ReactTree> => {
+    return (() => {
```
kmcq, line 35: What does wrapping this in a function do?

packages/open-truss/src/configuration/engine-v1.tsx
kmcq: I agree. I tried doing this and the types got upset 🤔

```tsx
  const renderFrames = (frames: FrameV1[]): Promise<ReactTree> => {
    const foo = Promise.all(frames.map(async ({ view, data, frames: subFrame }, i) => {
      const { component, props: viewProps } = view
      const Component = COMPONENTS[component]
      const props = {
        key: i,
        data,
        config,
        ...viewProps,
      }

      if (subFrame === undefined) {
        return await Component({ ...props })
      } else {
        const children = <>{renderFrames(subFrame)}</>
        return await Component({ ...props, children })
      }
    }))
    return foo
  }
```

> Type 'Promise<Element[]>' is not assignable to type 'Promise<ReactTree>'.

packages/open-truss/src/configuration/engine-v1.tsx
hktouw: Oh good idea. We can always add that back in whenever we're ready to use it. 👍 

```suggestion
```

packages/open-truss/src/configuration/engine-v1.tsx
hktouw: I had to do this to match the return type. `async` functions require the return type to be `Promise<>`. Before, the type was `Array<Promise<>>` which it did not like. Kind of awkward, but it seems to be working.

packages/open-truss/src/configuration/engine-v1.tsx
hktouw: Let me play around with this to see if I can get this to work.

hktouw: I think I got this working here https://github.com/open-truss/open-truss/pull/60/commits/2103216c2bc12ce7b764864c5605183906831c51

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -25,23 +29,35 @@ export interface WorkflowV1 {
...
+            data,
+            config,
+            ...viewProps,
+          }
+
+          if (subFrame === undefined) {
+            return await Component({ ...props })
+          } else {
+            const subFrames = (await renderFrames(subFrame)).map((child, k) => {
+              return <div key={k}>{child}</div>
```
kmcq, line 50: Did you add a div to add a key? To not add a div you can use the long form for a fragment: <React.fragment key={k}>...</React.fragment>

hktouw: Oh, good to know! I'll add that to my todo to fix.

PR #62, opened 2023-12-21
Move OT development to nextjs app

hktouw:
## Why

We want to mitigate some of the concerns @jonmagic [raised](https://github.com/open-truss/open-truss/pull/57#pullrequestreview-1789470067) around development now requiring running `npm run ot:setup` in the demo app to copy over changes.

## How

This PR proposes we move the majority of development to the `packages/open-truss/nextjs` app and use the demo-app primarily as a place to test the integration with "real" target applications. This obviates the need of having to run `npm run ot:setup` when working on libraries, pages, or apis. Now we only need to run this when testing the integration in the demo-app.

As one would expect, you can run `script/server` in the nextjs dir and that will start the nextjs app and tsc:watch service. This makes it easy to iterate on one's changes in the nextjs app for library or application code. This should pretty much be exactly the same dev experience as we have in the demo-app.

I updated `bin/setup` to copy over `components-scaffold` and `configs-scaffold` instead of the original ones. That means we should feel free to add anything we want to the `components` and `configs` directory and not worry about it getting copied over.

The easiest way to review this PR is commit by commit and by reading the README.

h/t to @kmcq for a [great PR](https://github.com/open-truss/open-truss/pull/59) that cleans up the demo-app. I branched this PR off of yours.

@kmcq @jonmagic What do ya'll think about this approach?


kmcq:
I think I'm good with this, but how about we say out loud what we're trying to do? Here's my list:

### 1. To help someone to start using OT, we want a way to quickly add files to their Next application.

These files are of two flavors:

1. `src/app/ot/`
    - Things that use OT that we think all OT applications might want, like a playground and a ui config builder (coming soon).
1. `src/open-truss/`
    - Scaffolding for where applications can put their components and configs, with an example of each.

### 2. To help someone see OT's benefits and capabilities, we want a way to showcase worthwhile examples.

Things like the trino client included in this PR. Examples should be all things from basic uses to more complex ones.

### 3. To help developers of OT, we want a working application to act as a test case.

A place to try out new config versions, make sure we don't break things, etc.

----

As a user of OT, I want my scaffolding to be very curated and have only basic functionality. I don't want any complex examples to be added to my project; they should only rely on the open-truss library.

As a developer of OT, I want freedom to use the demo app in lots of different ways, creating many test cases.

I think this PR gets at pretty much everything I want 👍 I like that we now have separate places for internal test cases and scaffolding that OT users will use. I'm not sure our examples are curated enough yet, but I think that can come with time and it's not a priority right now.

----

@hktouw does this all align with your goals? I haven't read the code too deeply yet but I can later

kmcq:
Thinking about this a bit more, I think I want scaffolding to be separate from examples and test cases. Right now having the two side by side in `packages/open-truss/nextjs` is confusing to me.

hktouw:
> @hktouw does this all align with your goals? 

Yes indeed! I think this approach gives us the most flexibility while minimizing tedious work. But definitely open to other thoughts!

>  I think I want scaffolding to be separate from examples and test cases.

Sounds good to me! What do you think about:

```
|── open-truss
│   ├── components
│   ├── configs
│   ├── scaffolding
│   │   ├── components
│   │   ├── configs
```

kmcq:
@hktouw maybe we should close this out? And depending on how you're feeling about it, should we delete `packages/open-truss/nextjs` for now?

hktouw:
Closing this out for now sounds good to me 👍

PR #63, opened 2023-12-23
Add /api/graphql routes to demo-app to start using apollo server

jonmagic:
## Why?

I would like to start demonstrating a non-Next.js way of using open-truss that leverages an apollo server backend for querying and mutating data.

My plan is to build on this PR and add a `uqiQuery` query to the graphql schema. The backend will use the uqi library and any clients we've created in the demo-app to retrieve and return (and/or maybe stream?) data to the client. Then I'll build a simple query UI where a user can submit a query and view the results.

I think this will get us closer to starting to use open-truss in a more meaningful way at the day job.

## How?

I followed these instructions: https://javascript.plainenglish.io/using-apollo-server-with-nextjs-13-app-dir-part-1-google-auth-with-nextauth-prisma-ca242c92599e

You can test out this PR by:

```
cd demo-app
npm install
script/server
```

Then visit http://localhost:3000/api/graphql

### DIFF
--- a/demo-app/package-lock.json
+++ b/demo-app/package-lock.json
@@ -8,7 +8,10 @@
       "name": "model-home",
       "version": "0.1.0",
       "dependencies": {
+        "@apollo/server": "^4.9.5",
+        "@as-integrations/next": "^3.0.0",
         "@open-truss/open-truss": "file:../packages/open-truss",
+        "graphql-tag": "^2.12.6",
         "next": "13.5.5",
         "react": "^18",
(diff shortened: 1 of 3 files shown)


kmcq (approved):
Nice and easy! What do you think of Next's `route.ts` file that exports http methods? Seems like a fun pattern to me so far

jonmagic:
> What do you think of Next's route.ts file that exports http methods? Seems like a fun pattern to me so far

yeah, me too :)

PR #64, opened 2023-12-23
Render configs from an endpoint

kmcq:
## Why?

I want to be able to render a config from an endpoint. To do this I've made the demo app have two playgrounds: one for RSC at `/ot/rsc-playground` and one for client components at `/ot/playground`. If we don't want to worry about RSC, we could just delete that one. FWIW I'll be developing in the non-RSC one.

## How?

This is kinda a big PR but hopefully it is logical to read commit by commit. In addition to adding the new playground, I updated the engine to support non-async components. Making a client component (by adding `use client`) would break but now it works. I'm pretty sure I didn't break RSC either but I'll do some more checking.

### DIFF
--- a/demo-app/src/app/ot/api/configs/[slug]/route.ts
+++ b/demo-app/src/app/ot/api/configs/[slug]/route.ts
@@ -1,0 +1,23 @@
+import { promises as fs } from 'fs'
+
+const CONFIG_DIR = './src/open-truss/configs/'
+
+export async function GET(
+  _request: Request,
+  { params: { slug } }: { params: { slug: string } }
+) {
+  const path = `${CONFIG_DIR}${slug}.yaml`
+  try {
+    const config = await fs.readFile(path, 'utf-8')
+
+    return Response.json({ config })
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
+      return new Response('Not Found', {
+        status: 404,
+      })
+    } else {
+      throw err
+    }
+  }
+}
--- a/demo-app/src/app/ot/rsc-playground/[slug]/page.tsx
+++ b/demo-app/src/app/ot/rsc-playground/[slug]/page.tsx
@@ -1,12 +1,11 @@
-'use client'
 import path from 'path'
-import RenderFromEndpoint from '@/components/RenderFromEndpoint'
+import RenderFromFile from '@/components/RenderFromFile'
 
 // TODO: Set COMPONENT_INDEX in application config and OT loads it?
 import * as COMPONENTS from '@/open-truss/components'
 
 // TODO: Get this path from application config and only need to pass in filename?
-const CONFIG_API = '/ot/api/configs/'
+const CONFIG_DIR = './src/open-truss/configs/'
 
 export default function Page({
   params: { slug },
@@ -20,11 +19,10 @@ export default function Page({
   return (
     <>
       <h1>{sanitizedSlug}</h1>
-      <RenderFromEndpoint
+      <RenderFromFile
         components={COMPONENTS}
-        url={`${CONFIG_API}${sanitizedSlug}`}
+        path={`${CONFIG_DIR}${sanitizedSlug}.yaml`}
       />
     </>
   )
 }
-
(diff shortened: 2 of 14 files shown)


kmcq:
> In addition to adding the new playground, I updated the engine to support non-async components. Making a client component (by adding `use client`) would break but now it works. I'm pretty sure I didn't break RSC either but I'll do some more checking.

Okay after I wrote this I ran `npm run lint` and prettier added back in some `async`s because the types said `Promise<...>` and those `async`s broke  components that `use client` because "async/await is not yet supported in Client Components, only Server Components". So, I removed all the `Promise` type declarations and made my `PageTitle` component be an async function. And, RSC still works! And `use client` components also still work!

@hktouw you added in all that promise stuff to "update the engine to support async components" in https://github.com/open-truss/open-truss/pull/60. Can you check that what I've done doesn't break what you were doing?

demo-app/src/app/ot/api/configs/[slug]/route.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export async function GET(
+  _request: Request,
+  { params: { slug } }: { params: { slug: string } },
+): Promise<Response> {
+  const path = `${CONFIG_DIR}${slug}.yaml`
+  try {
+    const config = await fs.readFile(path, 'utf-8')
+
+    return Response.json({ config })
```
Comment (jonmagic) [inline demo-app/src/app/ot/api/configs/[slug]/route.ts:13]: Ooh, I like this and will have to update my PR to do it this way.

demo-app/src/app/ot/rsc-playground/[slug]/page.tsx
```diff
@@ -0,0 +1,28 @@
+import path from 'path'
+import RenderFromFile from '@/components/RenderFromFile'
+
+// TODO: Set COMPONENT_INDEX in application config and OT loads it?
+import * as COMPONENTS from '@/open-truss/components'
+
+// TODO: Get this path from application config and only need to pass in filename?
+const CONFIG_DIR = './src/open-truss/configs/'
```
Comment (jonmagic) [inline demo-app/src/app/ot/rsc-playground/[slug]/page.tsx:8]: I don't have answers to these questions, but I like the in-line TODO.

demo-app/src/components/AvailableWorkflowsFromEndpoint.tsx
```diff
@@ -0,0 +1,47 @@
...
+      const response = await fetch(`/ot/api/configs/`)
+      const json = await response.json()
+      setConfigs(json.config)
+    }
+    setLoading(true)
+    fetchConfigs().catch((e) => {
+      setError(e as Error)
+    })
+    setLoading(false)
+  }, [])
```
jonmagic, line 20: Seems worthy of a hook. If not in this PR then another soon.

demo-app/src/components/RenderFromEndpoint.tsx
```diff
@@ -0,0 +1,51 @@
...
+} from '@open-truss/open-truss'
+
+export default function RenderFromEndpoint({
+  components = {},
+  url,
+}: {
+  components?: COMPONENTS
+  url: string
+}): JSX.Element {
+  // TODO: Use UQI's REST client once that exists?
```
jonmagic, line 15: Ooh yeah, I like it 👍🏻 I can work towards this.

jonmagic (approved):
I am curious about the async/promise changes as I was pretty sure things were unhappy before those changes when viewing the project in vscode with eslint, prettier, and whatever typescript checking it does enabled.

Other than that looks great! I'm not at my laptop right now but I can pull this branch later and see if it is upset or just works in vscode.

demo-app/src/components/RenderFromEndpoint.tsx
kmcq: Yes! Happy to collab too

kmcq:
Yes please lemme know if VS Code is upset 🙏

kmcq:
Thanks @jonmagic! I opened #67 to address this.

hktouw:
Really cool! Fantastic work! 🥳 🥳 🥳 

> I removed all the Promise type declarations and made my PageTitle component be an async function. And, RSC still works! And use client components also still work!
> @hktouw you added in all that promise stuff to "update the engine to support async components" in https://github.com/open-truss/open-truss/pull/60. Can you check that what I've done doesn't break what you were doing?

I have no idea how this is still working, but oh well! Let's run with this for now and we can make adjustments if necessary!

PR #67, opened 2023-12-24
Consolidate component getting and fix up types

kmcq:
## Why?

I noticed after merging #64 that the playground pages were upset with a type error because the `AsyncPageTitle` component doesn't adhere to OT's `BaseOpenTrussComponentV1` type, specifically that it is async and thus doesn't have a ReturnType of `JSX.Element` but instead of `Promise<JSX.Element>`.

## How?

Okay so maybe this is a smell, but I got around this by erasing the types of the imported components.

If we extend `BaseOpenTrussComponentV1` to return `JSX.Element | Promise<JSX.Element>`, then to make things happy we have to add an `async` to the engine, which breaks rendering client components.

### Considerations

I wonder if folks think this solution, erasing the types of the imported components, is unacceptable. If it is, perhaps now is the time to discuss dropping RSC or not.

I think there may be another solution in here somewhere that we could explore to use a generic type or an `allowAsync` argument to the engine and switching on one of those to do the async stuff or not.

### DIFF
--- a/demo-app/src/app/ot/playground/[slug]/page.tsx
+++ b/demo-app/src/app/ot/playground/[slug]/page.tsx
@@ -2,12 +2,6 @@
 import path from 'path'
 import RenderFromEndpoint from '@/components/RenderFromEndpoint'
 
-// TODO: Set COMPONENT_INDEX in application config and OT loads it?
-import * as COMPONENTS from '@/open-truss/components'
-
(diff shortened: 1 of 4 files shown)


jonmagic (approved):
I'm ok with the smell for now if it keeps us moving forward. We can revisit later.

Issue #58, opened 2023-12-19
[Task] Streamline running the demo-app first time


jonmagic:
1. There are no instructions for getting started with the demo app.
2. If you guess to `cd demo-app && npm install && script/server` you would be wrong because you first need to `cd packages/open-truss && npm install` and then go to the demo-app dir and do things.

:pear: with @tmetz

PR #68, opened 2023-12-24
Update readme to be accurate for how we're using the demo app

kmcq:
The demo app's README had fallen out of date for how we'll be building things in the demo app and not in `packages/open-truss/nextjs`.

Also, this might resolve https://github.com/open-truss/open-truss/issues/58 as I added a sentence about how you need to `npm i` in the OT package.


jonmagic (approved):
What do you think about adding a script/bootstrap with checksum and have that run first in script/server?

kmcq:
> What do you think about adding a script/bootstrap with checksum and have that run first in script/server?

I love it! But I don't have access to an example of this anymore, could you give me an example to work off of?

kmcq:
@jonmagic I updated this to reflect what you did in #74 🎉

PR #69, opened 2023-12-25
Convert all types to zod

kmcq:
## Why?

I want to create a UX for building configs and to do that it will be helpful to inspect a component's props at runtime.

## How?

[`zod`](https://zod.dev/) is a popular runtime validation library that also has great TypeScript support. In this PR I've converted all our types to use `zod`. Well, almost all, Yaml is super recursive which is difficult, but I got it handled!

**NOTE:** I didn't bump the engine version as there are no breaking changes here. In fact, there are no changes at all! The exported types should be equivalent.

In a subsequent PR I'll show you how we can inspect zod's prop descriptions at runtime and build a UX on top of it. I've already validated this so I think this approach is worthwhile. I want to merge this sooner than I'll get to the next part since it's a pain to update these types.

### DIFF
--- a/packages/open-truss/src/configuration/engine-v1.tsx
+++ b/packages/open-truss/src/configuration/engine-v1.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { YamlObjectShape, YamlShape } from '@/utils/yaml'
+import { YamlObjectShape, YamlShape } from '../utils/yaml'
 import { type RenderingEngine, type ReactTree, type COMPONENTS } from './apply'
 import { z } from 'zod'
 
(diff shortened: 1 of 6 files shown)


packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -1,33 +1,46 @@
...
-  key?: number
-  children?: Components
-  data: YamlType
-  config: WorkflowV1
+const FrameBase = z.object({
+  view: z.object({
+    component: z.string(),
+    props: YamlObjectShape,
+  }),
+  data: z.string(),
```
jonmagic, line 13: I think this will change to an object but don't know what the shape will be yet.

jonmagic (approved):
Love it! Merry Christmas!

packages/open-truss/src/configuration/engine-v1.tsx
kmcq: Oops yea this should be `YamlShape`! It was a string early on when I first started this but was fixed recently. I'll update then merge. Thanks for the review!

hktouw (commented):
Really cool! I'm glad this wasn't too heavy of a lift! This is a great base to start working off of when I try and implement the config spec to work with some existing workflows we have.

PR #70, opened 2023-12-26
describeZod creates a readable description of a zod type

kmcq:
## Why?

Continuing on from #69, I am on a journey to create a UX for building an OT config. This PR introduces `describeZod` which creates a readable description of a zod type. See the tests for what it creates!

## How?

Some trial and error of spelunking into how Zod works. This uses some internals of Zod, so is perhaps brittle because of that, but I think it's relatively safe to do and if it breaks it's not like all of OT would break, just the config builder.


jonmagic (approved):
Looks great 👍

packages/open-truss/src/utils/descibe-zod.ts
```diff
@@ -0,0 +1,55 @@
...
+// describeZod outputs a readable summary of a zod type.
+// In Open Truss we use this to describe a component's prop types
+// so that we can create a UX to guide filling those props in.
+export function describeZod(
+  zobject: object,
+): Record<string, ZodDescriptionObject> {
+  return Object.entries(zobject).reduce<Record<string, ZodDescriptionObject>>(
+    (description, [key, value]) => {
+      // We don't process the base props like config and data.
+      if (Object.keys(BaseOpenTrussComponentV1PropsShape.shape).includes(key)) {
```
hktouw, line 22: It does feel like `describe-zod` could be an isolated function that you pass in dependencies to like `BaseOpenTrussComponentV1PropsShape`. But I don't think that would get us much right now, so this looks 👍 to me.

hktouw (commented):
Really awesome!

PR #72, opened 2023-12-26
Fix issue where subframes weren't getting keys

kmcq:
## Why?

I was getting console warnings about components not having unique keys. It turns out that calling the component as a function like this breaks that. Since we don't need to `await` these anymore, we can just construct JSX per usual.

## How?

Please notice my branch name :)

### DIFF
--- a/packages/open-truss/src/configuration/engine-v1.tsx
+++ b/packages/open-truss/src/configuration/engine-v1.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
-import { YamlObjectShape, YamlShape } from '../utils/yaml'
-import { type RenderingEngine, type ReactTree, type COMPONENTS } from './apply'
 import { z } from 'zod'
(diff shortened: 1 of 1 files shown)


jonmagic (approved):
> Please notice my branch name :)

😂

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -62,15 +61,19 @@ export function engineV1(
       }
 
       if (subFrame === undefined) {
-        return Component({ ...props })
```
hktouw, line 65: If / when we do need to go back to async components, let's remember this is a valid way to call components and how to fix the bugs we were seeing.

hktouw (commented):
> Please notice my branch name :)

I laughed out loud and looked around the office to see if anyone heard me laughing

PR #74, opened 2023-12-26
Developer experience improvements

jonmagic:
## Why?

We noticed some rough edges while trying to move quickly the past week.

## How?

1. Added `.env.local.example`. Next.js automatically uses `.env.local` files for environment variables.
2. Fixed something in the `/api/synchronous-uqi-query` route that I meant to fix in a previous PR.
3. Added `script/bootstrap` that attempts to install dependencies and write/use a checksum to speed up future runs of `script/bootstrap`.
4. Updated `script/server` to run `script/bootstrap`.
5. Updated `script/bootstrap` to copy `.env.local.example` to `.env.local` if it does not exist.


demo-app/Brewfile
```diff
@@ -0,0 +1,2 @@
+brew "overmind"
```
kmcq, line 1: fwiw I don't think we're using `overmind`, but I would be down to do so!

demo-app/script/bootstrap
```diff
@@ -0,0 +1,82 @@
...
+demo_app_dir=$(dirname "$script_dir")
+root_dir=$(dirname "$demo_app_dir")
+open_truss_package_dir="$root_dir/packages/open-truss"
+
+set -e
+
+# md5sum function used for calculating the checksum.
+if ! which md5sum >/dev/null 2>/dev/null; then
+  md5sum() {
+    md5
```
kmcq, line 15: 👍 

```
❯ which md5
/sbin/md5
```

kmcq (approved):
Thanks!

demo-app/Brewfile
hktouw: 👍 to overmind, especially if we eventually want to setup background jobs.

PR #75, opened 2023-12-26
engine-v1: type improvements, handle named exports, catch misbehaving children

kmcq:
While working on my config builder UX I noticed a few things could be improved.

- Add OT_COMPONENTS as an export of all OT components.
- engine-v1 supports components that have named exports.
  - This is important as I want to introduce exporting a `Props` constant to inspect a component's prop types.
- Mark some optional props as optional.
- Throw an error when children aren't supported but are attempted to be used.

### DIFF
--- a/packages/open-truss/src/components/index.ts
+++ b/packages/open-truss/src/components/index.ts
@@ -1,1 +1,3 @@
-export { default as OTExampleComponent } from './OTExampleComponent'
+import * as _OTExampleComponent from './OTExampleComponent'
+export const OTExampleComponent = _OTExampleComponent
+export const OT_COMPONENTS = { OTExampleComponent }
--- a/packages/open-truss/src/configuration/engine-v1.tsx
+++ b/packages/open-truss/src/configuration/engine-v1.tsx
@@ -11,19 +11,23 @@ import {
 const DataShape = YamlShape.optional()
 
 const FrameBase = z.object({
+  frame: z.null(), // used only to make configs more readable
   view: z.object({
     component: z.string(),
-    props: YamlObjectShape,
+    props: YamlObjectShape.optional(),
   }),
   data: DataShape,
 })
 
-type FrameType = z.infer<typeof FrameBase> & {
-  frames: FrameType[]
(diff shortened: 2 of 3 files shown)


packages/open-truss/src/components/index.ts
```diff
@@ -1 +1,3 @@
-export { default as OTExampleComponent } from './OTExampleComponent'
+import * as _OTExampleComponent from './OTExampleComponent'
+export const OTExampleComponent = _OTExampleComponent
```
kmcq, line 2: This is a breaking change, as anyone who was importing this component directly would need to do `OTExampleComponent.default` now, but I'm not worried about it.

jonmagic (approved):
This is really nifty 😄

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -29,18 +38,43 @@ const WorkflowV1Shape = z.object({
...
+  | z.infer<typeof BaseOpenTrussComponentV1PropsShape>
+  | ComponentWithChildren
 
 export type BaseOpenTrussComponentV1 = (
-  props: z.infer<typeof BaseOpenTrussComponentV1PropsShape>,
-) => Components
+  props: BaseOpenTrussComponentV1Props,
+) => React.JSX.Element
+
+function hasDefaultExport(
```
hktouw, line 59: This 🤯  my 🧠

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -1,24 +1,33 @@
...
-type FrameType = z.infer<typeof FrameBase> & {
-  frames: FrameType[]
+export type FrameType = z.infer<typeof FrameBase> & {
+  frames?: FrameType[]
 }
 
 const FrameV1Shape: z.ZodType<FrameType> = FrameBase.extend({
-  frames: z.lazy(() => FrameV1Shape).array(),
+  frames: z
+    .lazy(() => FrameV1Shape)
```
hktouw, line 28: I'm glad you figured out how to do recursive types, since we love those so much!

PR #77, opened 2023-12-27
config builder: Add support for nesting child frames

kmcq:
## Why?

Nesting frames is important and the config builder should support it.

## How?

This was fun! I described my solution in a comment in the code:

```
// While iterating over the lines of the config, we construct what
// each 'frames:' line's `framePath` is. This will look something like
// 'workflow.frames.0.frames' where the integer is the index of the
// current frame in the current frames array. This index gets
// incremented when we encounter a '- frame:' line.
```

This allowed me to create a nice UX where the rendered config YAML can both be where users choose where to nest and tell you where we're currently nesting. Right now I'm just changing the color of `frames:` to pink 🦩

In this screenshot I am rendering each level's `thisFramesPath` inside the button that sets the current `framePath` to `thisFramePath`. This was helpful for me when working on this and I think helps show how it works so I kept it for this example:

Screenshot 2023-12-26 at 11 18 22 PM


demo-app/src/app/ot/config-builder/page.tsx
```diff
@@ -172,21 +175,30 @@ function ComponentListItem({
...
+    const frames = (get(parsedConfig, framesPath, []) as FrameType[]).concat(
+      frame,
+    )
     setConfig(
-      stringifyYaml({
-        ...parsedConfig,
-        workflow: { ...parsedConfig.workflow, frames },
-      }),
+      stringifyYaml(
+        set(parsedConfig, framesPath, frames) as unknown as YamlType,
```
kmcq, line 200: TIL about lodash's `set` and I love it 😻

demo-app/src/app/ot/config-builder/page.tsx
```diff
@@ -218,11 +228,81 @@ function ComponentList(): React.JSX.Element {
...
+                <button
+                  onClick={() => {
+                    setFramesPath(thisFramesPath)
+                  }}
+                >
+                  Nest at this level
+                </button>
+              </pre>
+            )
+          } else {
```
kmcq, line 290: Next up I'm going to look for `- frame:` lines and add a button to delete frames.

jonmagic (approved):
I like the direction the config builder is headed 🤘

demo-app/src/app/ot/config-builder/page.tsx
```diff
@@ -218,11 +228,81 @@ function ComponentList(): React.JSX.Element {
   )
 }
 
+const FRAMES_PATH_PARTS = ['workflow:', 'frames:', '- frame:']
+
+function ConfigYaml({ config }: { config: string }): React.JSX.Element {
```
hktouw: As always you are a master at query / string parsing lol.

demo-app/src/open-truss/components/YourAppExampleComponent.tsx
```diff
@@ -1,7 +1,16 @@
-import { type BaseOpenTrussComponentV1Props } from '@open-truss/open-truss'
+import {
+  withChildren,
+  BaseOpenTrussComponentV1PropsShape,
+} from '@open-truss/open-truss'
+import { z } from 'zod'
+
+export const Props = withChildren(BaseOpenTrussComponentV1PropsShape).extend({
```
hktouw, line 7: Very clean, I like it 👍

hktouw (commented):
Looks great! This is really starting to shape up quite nicely!

PR #84, opened 2024-01-11
Engine v1 improvements

hktouw:
## Why

In order to build generic, re-usable components we need to be thoughtful of what dependencies OT components have. For smaller dependencies, they can be added to the project. For larger ones (e.g. `react-router`, `next`), those should likely be abstracted away and provided by clients.

## How

This PR extends the rendering engine to support dependency injection for components. For example, this PR extracts the `next/link` dependency out of the `AvailableWorkflowsFromEndpoint` component and puts it in a new `NextLink` component. This allows us to make generic components and have their dependencies provided by clients.

For example, the OTAvailableWorkflowsFromEndpoint is declared in the OT package, but the NextLink component is declared in the demo app and passed in as a dependency

```yaml
view:
  component: OTAvailableWorkflowsFromEndpoint
  props:
    link:
      type: component
      value: NextLink
```

The easiest way to review this PR is commit by commit.

## Follow up work

- **Design code organization for rendering engine** - the engine code is a tad spaghetti right now. I plan on re-organizing it to make it easier to make modifications to and scale gracefully.
- **Make config validator** - It would be nice if we had validation system for the config before rendering the component tree. That would let us "statically" check configs and surface errors before runtime. This separation makes it easy to create tests, but I'm going to try and see if this can be done primarily using zod's type system / validation engine. I'll try and slot a space for this when I re-organize the code structure.
- **Support default props for components** - Having to specify dependencies every time a component is used it a little tedious. I'm going to add config support for specifying defaults. This will be interesting too because this is going to be a global config and not just for a given workflow.


demo-app/src/open-truss/configs/2-available-workflows.yaml
```diff
@@ -0,0 +1,10 @@
+workflow:
+  version: 1
+  frames:
+    - frame:
+      view:
+        component: AvailableWorkflowsFromEndpoint
```
hktouw, line 6: @jonmagic @kmcq when you have a moment I would appreciate your help debugging an issue I'm seeing.

`AvailableWorkflowsFromEndpoint` is declared in the demo-app and `OTAvailableWorkflowsFromEndpoint` is declared in the package. They are pretty much identical, but I'm running into a react error for the OT version. Ya'll are much better versed in React than me, so any help would be appreciated!

Model Home 2024-01-11 at 4 47 44 PM


オートメーション 2024-01-11 at 4 47 11 PM

jonmagic: NOTE: I merged a branch with some version changes into main and then merged main into your branch here.

I'm not able to reproduce this error, what url are you accessing? I tried the following urls and they load fine:

- http://localhost:3000/ot/playground
- http://localhost:3000/ot/playground/2-available-workflows

hktouw: Ah, you need to update the config to use `OTAvailableWorkflowsFromEndpoint` and not `AvailableWorkflowsFromEndpoint`. That will update it to use the component that is in the package instead of the demo-app.

jonmagic (approved):
Took me a bit to get through this and grok it but I did and I dig it 😄 Seeing it all come together [in this commit](https://github.com/open-truss/open-truss/pull/84/commits/06285d560e5d88f9b816fb3061ac52ec76b619ce) was 🤌🏻

jonmagic:
I created 3 issues from the Follow Up Work above:
- https://github.com/open-truss/open-truss/issues/91
- https://github.com/open-truss/open-truss/issues/92
- https://github.com/open-truss/open-truss/issues/93

PR #89, opened 2024-01-14
Fix how @open-truss/open-truss package uses react

jonmagic:
## Why?

> Invalid Hook Call
> Hooks can only be called inside the body of a function component.
> There are three common reasons you might be seeing it:
> 1. You might have mismatching versions of React and React DOM.
> 2. You might be breaking the [Rules of Hooks](https://legacy.reactjs.org/docs/hooks-rules.html).
> 3. You might have more than one copy of React in the same app.

## How?

Read through https://iws.io/2022/invalid-hook-multiple-react-instances a few times and then updated `script/bootstrap` to create the correct link so the open-truss package is linked to the react version the demo-app is using.

### DIFF
--- a/demo-app/script/bootstrap
+++ b/demo-app/script/bootstrap
@@ -69,7 +69,15 @@ else
 fi
 
 # Install Node.js dependencies.
-cd $open_truss_package_dir && npm install && cd $demo_app_dir && npm install
+cd $open_truss_package_dir \
+  && npm install \
+  && cd $demo_app_dir \
+  && npm install
+
+# Fix react link
+cd $open_truss_package_dir \
+  && npm link $demo_app_dir/node_modules/react \
+  && cd $demo_app_dir
 
 # Copy .env.local.example to .env.local if .env.local doesn't exist.
 if [ ! -f $demo_app_dir/.env.local ]; then


kmcq:
Thanks!

kmcq (approved):

hktouw (approved):
Great find! Thanks for fixing this!

I ran this locally and was surprised to see `/package/open-truss/package.json` not modified. The link has to be created somewhere so node knows to use the demo-app's react. I would have assumed the package's `package.json` is where the link would be created just like how demo-app's package.json references `file: ../packages/open-truss` for the open-truss package. 

After poking around a bit more it seems like the link is made by just making a symlink in the package's `node_modules` directory. That probably makes sense because that link is not something that you'd want checked in and distributed in the npm package.

1  tmux 2024-01-15 at 2 02 41 PM

This change in the bootstrap script is pretty invisible. While probably highly unlikely to cause issues, if one does run into an issue it could be pretty difficult to debug as there is a magic symlink in the `node_modules` directory. Not sure how / where this would be documented so maybe there is no action item here, but just something for us to be aware of?

jonmagic:
Yeah, I too was surprised there wasn't any changes needed to package.json or package-lock.json. 👍 for documenting and maybe refactoring so linking/unlinking is a separate script. I can see if you're only doing dev on the package and don't want to use the demo-app to drive dev that you wouldn't want to use the symlinked version.

PR #90, opened 2024-01-14
Add a TrinoDemo workflow

jonmagic:
## Why?

We want to show off an end to end workflow that fetches data and displays it.

## How?

Review commit by commit.

### DIFF
--- a/demo-app/src/lib/synchronous-uqi-query.ts
+++ b/demo-app/src/lib/synchronous-uqi-query.ts
@@ -1,0 +1,64 @@
+import sources from '@/lib/uqi-sources'
+import { type UqiMetadata } from '@open-truss/open-truss'
+
+interface SynchronousQueryResult {
+  rows: SynchronousQueryRow[]
+  metadata: SynchronousQueryMetadata
+}
+
+interface SynchronousQueryRow {
+  values: SynchronousQueryValue[]
+}
+
+interface SynchronousQueryValue {
+  key: string
+  type: string
+  value: string
+}
+
+interface SynchronousQueryMetadata {
+  columns: SynchronousQueryColumn[]
+}
+
+interface SynchronousQueryColumn {
+  name: string
+  type: string
+}
+
+interface Args {
+  source: string
+  query: string
+}
+
+async function synchronousUqiQuery(
+  _object: unknown,
+  { source, query }: Args,
+  _context: unknown,
+): Promise<SynchronousQueryResult> {
+  const { config, createClient } = sources[source as keyof typeof sources]
+  const client = await createClient(config)
+  const queryIterator = await client.query(query)
+  const rows = []
+  let metadata: UqiMetadata = {
+    columns: [],
+  }
+  for await (const { row, metadata: m } of queryIterator) {
+    if (metadata.columns.length === 0) {
+      metadata = m
+    }
+    const values = Object.entries(row).map(([key, value]) => ({
+      key,
+      // TODO: cache the type mapping?
+      type: m.columns.find((column) => column.name === key)?.type || 'unknown',
+      value: String(value),
+    }))
+    rows.push({ values })
+  }
+
+  return {
+    rows,
+    metadata,
+  }
+}
+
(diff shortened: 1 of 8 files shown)


packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -99,7 +102,11 @@ export function engineV1(
       }
 
       if (subFrame === undefined) {
-        return <Component key={i} {...props} />
+        if (data) {
```
hktouw, line 105: If I'm understanding this PR correctly, the changes make it so that if `data` is specified in the yaml the engine passes it to `/api/synchronous-uqi-query` to query results then renders the component with the data. Given the following yaml in your PR, it will send the trino query to `/api/synchronous-uqi-query` and render the component with the trino results returned.

```yaml
   - frame:
      data:
        query: select * from  tpch.sf1.customer limit 10
        source: trino-demo
      view:
        component: TrinoDemo
```

I think the engine / yaml annotation will likely need to be a little more flexible, but what you have here looks great to me to ship and we can make changes in a follow up PR 👍. 

A few thoughts about the `data` annotation in no particular order:
- I can see a world where pages have multiple frames that each have their own queries that could be run asynchronously. It would be great if we built in the ability to execute those queries asap so frames aren't serial / blocking
- I can also see a world where some frames may have dependencies on parent frames for data and we would need to support the ability to load them serially.
- As pages have many pieces of data, there likely will be many data variables. Making sure we handle namespacing and avoiding collisions will be an interesting problem to solve.
- These and many other wrinkles all don't have to be solved today and I'm sure we'll get to them eventually. 
- When we're a bit farther along it may be an interesting exercise to do a figjam brainstorming session to talk through more sophisticated aspects of the yaml rendering engine to help guide development.

packages/open-truss/src/configuration/engine-v1.tsx
```diff
@@ -78,6 +80,7 @@ function hasChildren(component: any): component is ComponentWithChildren {
 
 export function engineV1(
   COMPONENTS: COMPONENTS,
+  uqiSources: UqiSources,
```
hktouw, line 83: This doesn't appear to be used right now. Is this going to be the list of available backends?

packages/open-truss/src/configuration/apply.tsx
```diff
@@ -18,6 +18,7 @@ export interface OpenTrussComponentExports {
 export type COMPONENTS =
   | Record<string, OpenTrussComponent>
   | Record<string, OpenTrussComponentExports>
+export type UqiSources = Record<string, unknown>
```
hktouw, line 21: Should this type come from the UQI library?

demo-app/src/lib/synchronous-uqi-query.ts
```diff
@@ -0,0 +1,64 @@
+import sources from '@/lib/uqi-sources'
```
hktouw, line 1: This file doesn't appear to be used. Did you intend to add this? Maybe you wanted to change the api's functionality to use this lib?

hktouw (approved):
Looks great to me! Really coming along!

demo-app/src/lib/synchronous-uqi-query.ts
jonmagic: It's used on line 38: https://github.com/open-truss/open-truss/pull/90/files#diff-0d36a671fa5661d4608140df44061a3f01682b4f6943bc4eb5b16248db76cfd6R38

packages/open-truss/src/configuration/engine-v1.tsx
jonmagic: Agreed, things will get more complicated soon 😄

packages/open-truss/src/configuration/engine-v1.tsx
jonmagic: Good catch, cleaned up in 06f7583e466c977285b74cc0a447b434361a0802

kmcq:
So cool to see this!

Issue #92, opened 2024-01-15
[Task] Make config validator


jonmagic:
Extracted from https://github.com/open-truss/open-truss/pull/84

> It would be nice if we had validation system for the config before rendering the component tree. That would let us "statically" check configs and surface errors before runtime. This separation makes it easy to create tests, but I'm going to try and see if this can be done primarily using zod's type system / validation engine. I'll try and slot a space for this when I re-organize the code structure.

kmcq:
@hktouw I just came across this awesome code:

https://github.com/open-truss/open-truss/blob/4069594f7bda2463439b24af1f7f50a8d1acbdfd/packages/open-truss/src/configuration/engine-v1/config-schemas.tsx#L56-L76

so rad! I think that where we're currently using it is too high up in the rendering process though:

https://github.com/open-truss/open-truss/blob/4069594f7bda2463439b24af1f7f50a8d1acbdfd/packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx#L25-L30

Right now, it fails the entire workflow if any part of the config is invalid. For workflow-level failures this makes sense. But for things that are incorrect inside a frame, I think instead it'd be better if we validate at the frame level. This way, frames without issue could render and those with bad configuration could fail and [be replaced with helpful error context UI](https://github.com/open-truss/open-truss/issues/103). What do you think? At the frame level is likely also a good place for [data errors](https://github.com/open-truss/open-truss/issues/80) to be caught.

hktouw:
@kmcq @jonmagic I think we pretty much have the config validator code in place. Now that is left is to better surface this validation in the UI. Perhaps we can add a step to the config builder that will run the zod parser. What do you all think?

kmcq:
I like that. But also what should happen if you try to render a bad config?

hktouw:
In the config builder, maybe we show the user any exceptions / zod parsing errors (but prettified to make it easier to read)? In production, perhaps the same?

Issue #91, opened 2024-01-15
[Task] Design code organization for rendering engine


jonmagic:
Extracted from https://github.com/open-truss/open-truss/pull/84

> The engine code is a tad spaghetti right now. I plan on re-organizing it to make it easier to make modifications to and scale gracefully.

hktouw:
Shipped this so I'm going to go ahead and close this issue

PR #94, opened 2024-01-17
Refactor engine code

hktouw:
ref https://github.com/open-truss/open-truss/issues/91
ref https://github.com/open-truss/open-truss/issues/92

### Why

To make iterating on the workflow engine faster and more scalable, we should improve the design of our code and file structure.

### How

This PR does a significant refactor of the engine code. This PR:
- Changes the file structure
- Moves around code to improve better organization and locality of logic
- Adds a config validation step before rendering

The easiest way to review this PR is to:
- Start with reading the `packages/open-truss/src/configuration/engine-v1/README.md` to get a sense for the new file structure and how to use the system.
- Pull down the code and start with the `engineV1` function in `packages/open-truss/src/configuration/engine-v1/engine.tsx`. `engineV1` is the main entry point and you can follow the code path to understand the new engine.

### Follow up

I can make issues for the following after folks have a chance to 👀 

- **Component level types and validations** - Now that we have a new validation engine setup, we can now setup the engine to at "config compile time" validate that the arguments passed to components adhere to the prop schemas the components have defined. This get's exciting when we apps create "complex types" which components can declare as requirements. For example, we can make components that declare a "User type" as a prop and validate at config compile type / runtime that that type of object is correctly being passed to the component. Complex types also make it possible for us to surface in the config builder all possible valid components that can be used as an argument for another component.
- **Config testing framework** - Now that we have validations setup, it should be easy to setup a testing framework where passing the engine a given config (with a bug), should trigger a validation error. That will allow us to build a config test suite we can run to ensure our engine is behaving as expected and prevent regressions.


jonmagic:
While testing we found out that two of the files in the refactor have a circular dependency. We moved the contents of `component.tsx` into `config-schemas.tsx` and got to the next error which was our demo workflows not rendering because they did not have a valid configuration.

packages/open-truss/src/configuration/engine-v1/renderer.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { type GlobalContext } from './engine'
+
+interface FrameContext {
+  frame: FrameV1
+  globalContext: GlobalContext
+  renderFrames: (frames: FrameV1[]) => ReactTree
+  i: number
+}
+
+export function renderFrame(
```
kmcq, line 25: First, this refactor is dope and made things much more readable.

Due to its readability, I now see a further enhancement we can make -- should `renderFrame` be a `<Frame />` component? It's a function that returns JSX, aka a component, so I think yes. And given that, I wonder if the whole rendering pipeline is actually just components, as what they do is render things and in React that's typically done by components.

I wonder if what this would give OT is a an export like `RenderConfig` that takes in a config prop and does all the stuff the engine is doing like checking `version` to see how to parse the config, etc. That could be nice because right now we're doing that lke this in the demo app and it's not beautiful:

```jsx
export default function RenderConfig({
  config,
}: {
  config: string
}): JSX.Element {
  const configurationFunction = React.useMemo(
    () => applyConfiguration(components),
    [],
  )

  const renderedComponents = React.useMemo(() => {
    if (config) {
      const parsedConfig = parseYaml(config)
      return configurationFunction(parsedConfig, {})
    }
  }, [config])

  return <>{renderedComponents}</>
}
```

I'd be happy to try to do this conversion; perhaps this weekend.

hktouw: > should renderFrame be a <Frame /> component?

Conceptually that makes sense to me and allows us to hide away the implementation detail bits like you referenced above. Let's give it a try!

> I'd be happy to try to do this conversion; perhaps this weekend.

Sounds good! I'll clean up this PR / fix the bugs. How do we feel about shipping this as is and you can do the conversion in a follow up PR?

After this PR, I'll continue build out features in the engine.

kmcq: I already did it 😊 https://github.com/open-truss/open-truss/pull/95

kmcq:
@hktouw when I try this branch out locally I get an error that `withChildrenV1` is not a function. Have you seen that?

jonmagic:
Delete the packages/open-truss/dist folder and restart the server.

hktouw:
This is ready for another 👀 and then we can merge it

demo-app/src/open-truss/components/NextLink.tsx
```diff
@@ -1,11 +1,11 @@
 import {
-  withChildren,
+  withChildrenV1,
```
kmcq, line 2: fwiw I don't love versioning `withChildren`. It's such a fundamental piece of React I can't imagine it changing

kmcq (approved):
This is great! (thanks @jonmagic for the tip to get it running)

I'm not sure I'm into the `V#` suffix anymore -- it's kinda ugly? I wonder if it'd be possible do something like (using the component version that #95 introduces) `import { RenderConfig } from open-truss/engine-v1` so the version is in the file you import from not the object itself. This doesn't block this PR.

EDIT: After reading #96 maybe actually we can keep the versioned components etc (although I still want `withChildren` to not be versioned) and the exported OT components like `RenderConfig`, `RenderFromEndpoint` etc handle any version switching inside of them?

demo-app/src/open-truss/components/NextLink.tsx
jonmagic: Done in [c3f6e98](https://github.com/open-truss/open-truss/pull/94/commits/c3f6e98ed59d62172e0a2b7144c9eea5e2799ffd)

hktouw:
> although I still want withChildren to not be versioned) and the exported OT components like RenderConfig, RenderFromEndpoint etc handle any version switching inside of them?

Seems reasonable to me!

PR #95, opened 2024-01-18
Refactor rendering engine to be components

kmcq:
As I talked about in https://github.com/open-truss/open-truss/pull/94#discussion_r1456736545, I think all of the rendering code fits better to be defined as components. This PR does it!

PR #96, opened 2024-01-18
Make client only package of OT and move Playground and Config builder there

hktouw:
@kmcq's [PR](https://github.com/open-truss/open-truss/pull/95) got me thinking that we can just make a client only version of the OT package and move all the `/ot` namespaces pages there. This seemed to work and didn't require any changes to the build processes. Now we can import these pages easily into our application.

`client.ts` - I created a new root level `client.ts` entry point file in the OT package that sits next to `index.ts`. I then modified our `package.json` to specify a new import namespace that points to `client.ts`. When you call `import blah from '@open-truss/open-truss/client` it uses `client.ts` as the root file.

This separate entry point file was necessary because when I initially moved the config-builder to the OT package, the server-side code `/ot/api/synchronous-uqi-query` was complaining that it was importing React's `useState`. This tells me we'll likely encounter this issue sooner than later for the OT components in the package, so we should probably move those to the client package eventually. And to that end, perhaps we probably should make Playground and ConfigBuilder into OT components 🤔 and not have the concept of `pages` quite yet.


demo-app/src/pages/ot/config-builder/index.tsx
```diff
@@ -1,35 +1,6 @@
...
-          <ComponentList />
-          <h2>config</h2>
-          <ConfigYaml config={config} />
-        </div>
-        <div>{config !== CONFIG_BASE && <RenderConfig config={config} />}</div>
-      </div>
-    </ConfigBuilderContext.Provider>
-  )
+export default function ConfigBuilder(): React.JSX.Element {
+  return <ConfigBuilderPage components={OT_COMPONENTS} />
```
kmcq, line 5: Yes! I love how this is a NextJS page but OT itself is agnostic and could be rendered using any framework ✨ 

I'm not sure why but the `*Page` suffix isn't my favorite. I'll noodle on that

demo-app/src/pages/ot/config-builder/index.tsx
```diff
@@ -1,35 +1,6 @@
-import RenderConfig from '@/components/RenderConfig'
-import { useState } from 'react'
-import ComponentList from './ComponentList'
-import ConfigYaml from './ConfigYaml'
-import {
-  CONFIG_BASE,
-  ConfigBuilderContext,
-  INITIAL_FRAMES_PATH,
-} from './config-builder-context'
+import { ConfigBuilderPage } from '@open-truss/open-truss/client'
```
kmcq, line 1: I'm loving that we're adopting specialized packages! (the `/client` stuff, dunno what that's called) I think I would suggest we go further though and put all of config builder in `/config-builder` and the playground in `/playground`. They feel optional, and to me that feels like folks should reach into specialized packages to get their contents. What do you think?

hktouw: I think a trade-off of moving pages to top-level imports is we could end up with a lot of top-level imports. Some pages that come to mind that we're building / will build are:
- config builder
- playground
- query editor
- workflow page (tbd)
- workflow session page (tbd)

Maybe instead of `/clients`, this group of features can be `/pages`?

hktouw: Changed `/client` to `/pages`. Let me know if that isn't the route we want to go and I'm happy to revert it https://github.com/open-truss/open-truss/pull/96/commits/e62690b1dff925dc579a7d795c196d5970f1f488

kmcq: That rename makes me feel good!

kmcq (approved):
I think this is a great direction.

It looks like the changes from #94 to how props get defined in the config (with a `type` and `value`) broke the config builder, which now shows up in the `tsc watch` output because it's part of the package. That's not related to this PR though so we can patch it after this.

hktouw:
> It looks like the changes from https://github.com/open-truss/open-truss/pull/94 to how props get defined in the config (with a type and value) broke the config builder, which now shows up in the tsc watch output because it's part of the package.

Really strange. The error we were seeing past the CI build for tests, but failed the one running on main that uploads packages to npm. https://github.com/open-truss/open-truss/actions/runs/7623115580. 

I can fix this bug tomorrow and also fix the build so that it catches these types of errors on PRs before they get to main.

PR #98, opened 2024-01-29
Instead of saying what type a prop is, use <Component /> syntax

kmcq:
This is an alternative solution to the `props[name].type`/`props[name].value` solution from https://github.com/open-truss/open-truss/pull/94: instead of saying that something has `type: component`, I am introducing a syntax way to specify it:

```diff
workflow:
  version: 1
  frames:
    - frame:
      view:
        component: OTAvailableWorkflowsFromEndpoint
        props:
          link:
-           type: component
-           value: NextLink
+         link: <NextLink />
```

While more magical, I like that this removes the need to specify that every string is a string and number is a number; this could really add up to a lot of lines in large configs.

What do you all think?

### DIFF
--- a/demo-app/src/pages/ot/playground/index.tsx
+++ b/demo-app/src/pages/ot/playground/index.tsx
@@ -9,9 +9,7 @@ workflow:
       view:
         component: OTAvailableWorkflowsFromEndpoint
         props:
-          link:
-            type: component
-            value: NextLink
+          link: <NextLink />
 `
 
 export default function Playground(): JSX.Element {
--- a/packages/open-truss/package.json
+++ b/packages/open-truss/package.json
@@ -53,6 +53,7 @@
   },
   "devDependencies": {
     "@types/jest": "^29.5.8",
+    "@types/lodash": "^4.14.202",
     "@types/react": "^18.2.0",
     "@typescript-eslint/eslint-plugin": "^6.9.1",
     "@typescript-eslint/parser": "^6.9.1",
--- a/demo-app/package-lock.json
+++ b/demo-app/package-lock.json
@@ -47,6 +47,7 @@
       },
       "devDependencies": {
         "@types/jest": "^29.5.8",
+        "@types/lodash": "^4.14.202",
         "@types/react": "^18.2.0",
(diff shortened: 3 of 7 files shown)


packages/open-truss/package.json
```diff
@@ -53,6 +53,7 @@
   },
   "devDependencies": {
     "@types/jest": "^29.5.8",
+    "@types/lodash": "^4.14.202",
```
kmcq, line 56: vim told me to do this and I'm not one to say no to vim

kmcq:
@hktouw this fixes the config builder too fyi so if we like this, we're all set there

hktouw (approved):
Hmmm, part of me _feels_ that we will want props to be typed, but I can't quite articulate why that is. 🤔 

I have an idea to do prop / data type checking by using the zod types declared by the component. These should even handle user-defined, custom types (e.g. User). In that world we probably don't need schema-level type checking and can probably delete it like you're doing. We lose some expressiveness in the config, but maybe that's ok?

demo-app/src/pages/ot/playground/index.tsx
```diff
@@ -9,9 +9,7 @@ workflow:
       view:
         component: OTAvailableWorkflowsFromEndpoint
         props:
-          link:
-            type: component
```
hktouw, line 13: What other types can you envision in the future? Some that come to mind are
- enum
- function
- component with args

Part of me is a bit concerned that by removing the type, it makes it a bit more fragile as we have to parse values to know what type it is.

kmcq: Of those three I don't know how to handle functions, but enums are just strings and a component with args should probably be done as children with sub frames.

I think I agree with you conceptually: giving up types feels a bit wrong. But in practice I'm not sure we've given much or anything up. Asking if something has `type === 'string'` is the same work as `typeof prop === 'string'`. Yaml natively supports strings, numbers (ints, floats), dates, and booleans so we can use standard `typeof`.

But again I don't know how to handle functions yet and that will be important. For now I'd be down to try it this way and see if it works with the zod stuff you're going to add in.

Issue #97, opened 2024-01-28
[Task] A way to inject a `<FrameWrapper />` that the engine wraps each frame with


kmcq:
I want the config builder to have an interface to edit a workflow directly on that workflow. Thinking through this, I think modifying `RenderFrame` take in a wrapper component (I like the name `FrameWrapper`) that the config building will set up to have things like deleting and editing frames will solve this nicely.

This also feels similar to how we set things up with `DataProvider` and I wonder if that will make this a small change. Of course `FrameWrapper` will be optional and nothing will be wrapped by default.

PR #101, opened 2024-02-01
Add the ability to configure a FrameWrapper

kmcq:
## Why?

Closes https://github.com/open-truss/open-truss/issues/97.

This will definitely be helpful for the config builder, but perhaps in other contexts as well. For the config builder, we can use this to have an interface to edit a workflow directly on that workflow

## How?

- Turn off react/prop-types and use typescript errors instead
- Add ability to wrap frames with a custom component
- Showcase FrameWrapper by adding one to the config builder

EDIT: I had a simpler example FrameWrapper but I've built a cooler one. You can now edit the page through buttons that the FrameWrapper renders on top of components 😎 

https://github.com/open-truss/open-truss/assets/3272924/8d5aee66-1be6-4cd0-9f0f-9d8d2d66f886

Here's the old video:

https://github.com/open-truss/open-truss/assets/3272924/14be460e-35a1-4b09-a4fd-21a1c6e6b62c


packages/open-truss/src/pages/config-builder-page/index.tsx
```diff
@@ -12,6 +16,24 @@ interface ConfigBuilderPageInterface {
   components: COMPONENTS
 }
 
+const Wrapper: FrameWrapper = ({ children, configPath, frame }) => {
```
kmcq, line 19: The eslint rule change I made makes declaring a function like this possible. Before you'd have to have a type for each prop defined but they're already defined in `FrameWrapper`

kmcq: I'll clean up other declarations like this if we're into it!

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -28,13 +31,26 @@ export function RenderConfig({
...
+    FrameWrapper,
+  }
 
   return (
     <>
-      {config.frames.map((frame, i) => (
-        <Frame key={i} frame={frame} globalContext={globalContext} />
-      ))}
+      {config.frames.map((frame, i) => {
+        const configPath = `workflow.frames.${i}`
```
kmcq, line 43: Having the the path to this frame in the config I think could be helpful for debugging. For example, we could tell you where in the config a component that is not configured correctly is.

packages/open-truss/src/pages/config-builder-page/index.tsx
hktouw: oh rad, that's way cleaner! yes please!

packages/open-truss/src/configuration/RenderConfig.tsx
```diff
@@ -19,19 +20,30 @@ export interface OpenTrussComponentExports {
...
+>
 
 export function RenderConfig({
   components: appComponents,
   config,
+  FrameWrapper,
 }: {
   components: COMPONENTS
   config: string
+  FrameWrapper?: FrameWrapper
```
hktouw, line 34: What do you think about making FrameWrapper into an OT component, adding a frame wrapper annotation to the config, and teaching the renderer to handle this? 

Over time I get the feeling that we'll be adding a lot of options to the renderer and instead of passing them through like we're doing here I wonder if we can extend the config to handle this. For now, this perhaps could be at the workflow level such as:

```
workflow:
  version: 1
  frameWrapper: ConfigBuilderFrameWrapper
  frames:
...
```

Maybe eventually we can create a global OT config where this could be set globally and perhaps group level configs. But a problem for another time :).

hktouw (commented):
Wooooow. This is soooo cool. Fantastic work Keith! 🎉 

> This will definitely be helpful for the config builder, but perhaps in other contexts as well.

I feel like this will be pretty important. In addition to what you've built for config builder which can be used for other views (e.g. forking workflows), I think we want the ability to mixin / inherit functionality across OT components and this could provide a way to do that. For example, if / when we get around to showing / hiding data based on permissions it would be great if we can apply that to all OT display components (of course this needs to happen with RSC so data never gets sent to the client).

packages/open-truss/src/configuration/RenderConfig.tsx
kmcq: whoa yes this could be cool and really powerful! Lemme take a look

kmcq: done in https://github.com/open-truss/open-truss/pull/101/commits/18ae1f0af75462d3f7779a85df2338fe9ae446cb 🎉 this is very cool I think

packages/open-truss/src/pages/config-builder-page/config-builder-context.tsx
```diff
@@ -0,0 +1,88 @@
+import get from 'lodash/get'
+import set from 'lodash/set'
+import { createContext, useState } from 'react'
+import { type FrameType, type WorkflowSpec } from '../../configuration'
+import { type YamlType, parseYaml, stringifyYaml } from '../../utils/yaml'
+
+export const CONFIG_BASE = `
+workflow:
+  version: 1
+  frameWrapper: ConfigBuilderFrameWrapper
```
kmcq, line 10: @hktouw now this is where this is configured!

packages/open-truss/src/pages/config-builder-page/index.tsx
```diff
@@ -1,27 +1,118 @@
...
+      {showConfig && (
+        <div>
+          <h2>config</h2>
+          <ConfigYaml />
+        </div>
+      )}
+      {!showConfig && (
+        <RenderConfig
+          config={config}
+          components={{ ...components, ConfigBuilderFrameWrapper }}
```
kmcq, line 104: I do this in two places in this PR now and I'm curious what you all think. This injects a component when we need it, making it private. I didn't think this and `OTDefaultFrameWrapper` needed to be exposed to package users.

jonmagic (approved):
Kicked the tires and this is so darn cool :ship:

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -28,13 +35,30 @@ export function RenderConfig({
...
+    COMPONENTS,
+    FrameWrapper,
+  }
 
   return (
     <>
-      {config.frames.map((frame, i) => (
-        <Frame key={i} frame={frame} globalContext={globalContext} />
-      ))}
+      {config.frames.map((frame, i) => {
```
hktouw, line 50: Hmm, this code is largely the same as the code in `Frame`. I wonder if there is a way to cast `config` as `frame` so we can just render the `Frame` and use much of the same code. 🤔

packages/open-truss/src/pages/config-builder-page/config-builder-context.tsx
hktouw: this is so cool 🚀

packages/open-truss/src/pages/config-builder-page/config-builder-context.tsx
```diff
@@ -0,0 +1,88 @@
...
+  addFrame: (_f: FrameType) => null,
+  deleteFrame: (_c?: string) => null,
+})
+
+export const Provider: React.FC<React.PropsWithChildren> = ({ children }) => {
+  const [config, setConfig] = useState<string>(CONFIG_BASE)
+  const [framesPath, setFramesPath] = useState<string>(INITIAL_FRAMES_PATH)
+
+  const addFrame = (frame: FrameType): void => {
+    const parsedConfig = parseYaml(config) as unknown as WorkflowSpec
```
hktouw, line 37: Do you think there is any utility to eventually make this parsed by zod? That would run validations and give us the type as opposed to us having to cast to `unknown` then `WorkflowSpec`. If we don't get much from this here though, then probably not worth the effort.

packages/open-truss/src/pages/config-builder-page/index.tsx
```diff
@@ -1,27 +1,118 @@
...
   ConfigBuilderContext,
+  Provider as ConfigBuilderContextProvider,
   INITIAL_FRAMES_PATH,
 } from './config-builder-context'
 
 interface ConfigBuilderPageInterface {
   components: COMPONENTS
 }
 
+const ConfigBuilderFrameWrapper: FrameWrapper = ({
```
hktouw, line 19: So cool :)

hktouw (commented):
Right on, this is fantastic. Nice work!

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
kmcq: Yes I think so! This exact thing is in my todos

packages/open-truss/src/pages/config-builder-page/config-builder-context.tsx
kmcq: Interesting! I think not yet because the config builder constructs the frame for you from the zod description. However, I was thinking we probably want to support editing the config directly by typing in the YAML and in that case doing this would be helpful

Issue #103, opened 2024-02-02
[task] Create an `<ErrorBoundary />` that replaces a failing frame with helpful error context


kmcq:
Per usual, I am motivated by the config builder. I envision folks being able to edit frames inline and have anything that is wrong with them be shown to the user instead of crashing the entire page.

kmcq:
I've been playing around with https://github.com/bvaughn/react-error-boundary/ and it's pretty legit

kmcq:
While working on https://github.com/open-truss/open-truss/pull/104 I decided to not use an error boundary for config errors and instead wrap some code with `try ... catch` and use that to render an error component. In dev mode, NextJS will always render the big error modal and I found that to be annoying. I also think perhaps errors around frame validation shouldn't throw errors and instead could construct error messages. I didn't do the work to migrate `getComponent` to not `throw` but we should probably move in that direction.

PR #104, opened 2024-02-03
Catch and display frame configuration errors

kmcq:
Closes https://github.com/open-truss/open-truss/issues/103.

While building a config, having the page break totally when you misconfigure something is an unpleasant experience. I've taught `<Frame />` to catch errors related to failing to find a component and render a UI saying what's up:

image

This could be where we do other frame-level config validations, like prop validation.

### DIFF
--- a/packages/open-truss/src/configuration/engine-v1/Frame.tsx
+++ b/packages/open-truss/src/configuration/engine-v1/Frame.tsx
@@ -17,6 +17,33 @@ interface FrameContext {
   configPath: string
 }
 
+class FrameError extends Error {
+  public componentName: string
+  public configPath: string
+
+  constructor(message: string, componentName: string, configPath: string) {
+    super(message)
+    this.componentName = componentName
+    this.configPath = configPath
+  }
+}
+
+function ShowError({ error }: { error: FrameError }): JSX.Element {
+  return (
+    <div
+      role="alert"
+      style={{ border: '1px solid #ececec', borderRadius: '5px' }}
+    >
+      <pre style={{ color: '#777', margin: 0 }}>
+        {error.componentName}({error.configPath})
+      </pre>
+      <div style={{ padding: '5px' }}>
+        <pre style={{ color: 'red', textAlign: 'center' }}>{error.message}</pre>
+      </div>
+    </div>
+  )
+}
+
 export function Frame(props: FrameContext): React.JSX.Element {
   const {
     frame: { view, data, frames },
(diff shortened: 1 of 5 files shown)


packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -27,14 +27,14 @@ export function RenderConfig({
...
-    // TODO for now just raise any config validation errors.
-    // We likely want to render something nicer eventually.
-    throw result.error
-  }
+  // const result = WorkflowV1Shape.safeParse(config)
+  // if (!result.success) {
+  //   // TODO for now just raise any config validation errors.
+  //   // We likely want to render something nicer eventually.
+  //   throw result.error
+  // }
```
kmcq, line 35: Like I said in https://github.com/open-truss/open-truss/issues/92#issuecomment-1922844648, I think it would be best to move most config validation to the frame level so we can do things like I've done this in PR. Would love to hear what you all think!

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -94,20 +133,29 @@ function processProps({
...
 ): OpenTrussComponent {
   const componentName = component.replaceAll(/(<|\/>)/g, '').trim()
   let Component = COMPONENTS[componentName]
   if (!Component) {
-    throw new Error(`No component '${componentName}' configured.`)
+    throw new FrameError(
+      `No component '${componentName}' configured.`,
+      componentName,
+      configPath,
+    )
```
kmcq, line 146: x-posted from https://github.com/open-truss/open-truss/issues/103#issuecomment-1925465268:

> While working on https://github.com/open-truss/open-truss/pull/104 I decided to not use an error boundary for config errors and instead wrap some code with `try ... catch` and use that to render an error component. In dev mode, NextJS will always render the big error modal and I found that to be annoying. I also think perhaps errors around frame validation shouldn't throw errors and instead could construct error messages. I didn't do the work to migrate `getComponent` to not `throw` but we should probably move in that direction.

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
hktouw: Oh interesting. So you're thinking at every `Frame` we call `FrameV1Shape.safeParse(config)` on the current node of the config and render the errors accordingly?

kmcq: I think so, yea! There might be a top level validation too, but in my opinion it shouldn't validate any frames.

hktouw: I think a pro / con might be
- pro 
  - will have `configPath` and can show exact config that failed
  - can render an error right on the frame in question
- con
  - we no longer will be front-loading validation. Maybe we can still do this for production and smoke tests, but do this PR's approach for config builder like tools

hktouw (approved):
Looks good to me! Let's give this a go!

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
kmcq: Interesting con there! I think this PR could mostly stay the same but I'll:

1. Make the top-level validation always happen unless you tell it not to with a prop
2. Clean up how errors are thrown a bit so `configPath` doesn't need to be set at the error. This will let this system catch any errors that occur, not just configuration ones.

PR #106, opened 2024-02-07
Add more validations for props

### LINKED ISSUE #105: Validate that props match what components expect
Thanks to Zod and Keith's great idea to have components declare what props they take, we have the ability at runtime (or config compile time) to know what values are valid for a given component. We should update our zod schemas to perform validation checks on props to assert they match the types and values that the components expect. Later, we can also apply this to data passed to components.

hktouw:
resolves https://github.com/open-truss/open-truss/issues/105

## Why

We want more prop level validations to ensure configs are valid

## How

This PR does a little bit of refactoring, but the bulk of it adds more validations. Notably the following View-level validations:
  - validate component declared in the config exists
  - validate props match what component expects
    - validate scalars are the correct type
    - validate components passed as props exist
    - validate enums and ensure value passed in is valid

This should be relatively straightforward to review. You can review this commit by commit or by reading through the changes in the config-schemas file and jumping to refactors that support the changes.

### DIFF
--- a/packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
+++ b/packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
@@ -2,6 +2,9 @@ import { z } from 'zod'
 import { YamlShape } from '../../utils/yaml'
 import { type OpenTrussComponentExports } from '../RenderConfig'
 import { RUNTIME_COMPONENTS } from './RenderConfig'
+import { parseComponentName } from './Frame'
+import { describeZod } from '../../utils/descibe-zod'
+import { type ZodDescriptionObject } from '../../utils/descibe-zod'
 
 // Data Schemas
 /*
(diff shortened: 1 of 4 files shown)


packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
```diff
@@ -28,6 +31,47 @@ workflow:
...
+    if (view.props && hasPropsExport(Component)) {
+      const expectedComponentProps = describeZod(Component.Props.shape)
+      Object.entries(view.props).forEach(([propName, propValue]) => {
+        const expectedProp = expectedComponentProps[propName]
+        if (expectedProp === undefined) {
+          // TODO probably fine to allow props not declared by the component?
+          return
+        }
+
+        const errors = isPropValueValid(expectedProp, propValue)
```
kmcq, line 62: I'm a little surprised we need this -- `isPropValueValid` looks a lot like reimplementing zod's validation system. Can you say more about why you did it this way versus zod's validations, which I think would be `Component.Props.parse(view.props)`?

PR #110, opened 2024-02-11
[Spike] State management

hktouw:
## Why

To support more complex workflows where:
- components are sharing and passing data to other components
- we programmatically show what workflows can be linked with other workflows based on their interface

We likely need a state management solution.

## How

This PR spikes out a state management framework using the preact signals library. In the demo below, you can see two separate components reading and modifying the same state.

https://github.com/open-truss/open-truss/assets/4481584/6e1fb81f-6bd0-4192-aa84-78c7d4af615c

### Implementation

At a high level, this PR
- Adds a new "Signal\<type\>" type that can be a valid prop for open truss components.
- Modifies the renderer to check all component props for a given workflow. If a prop is of type "Signal", the renderer creates a new preact signal of that type and saves into the workflow's globalContext object. When we render the component, we grab the signals it needs from the globalContext and pass it to the component. This allows components to share the same signals and read and update them globally.

With the addition of signal types, in the future this should make it easier to dynamically show which components can be linked with other components by virtue of sharing the same props interface. e.g. a form component that modifies the AccountIDs signal can be linked with a display component that reads an AccountIDs signal.

### Considerations

- One thing I'm not the biggest fan of with this approach is that it is very magical. None of this functionality is displayed in the yaml config. That makes configs easier to write, but makes it a bit harder to reason about. If we go this route, I suspect we'll need to create debugging tools such as
  - showing what the values in the global state store
  - showing which components are using which signals
  - making the config builder really user friendly for making sure the workflow is doing what you want
- I did this spike in the context of trying to figure out how to support complex multi page workflows. e.g. Page 1: show form, Page 2: display results based on that form. If we want to make these steps composable with different components, it seemed like we'll need some system to share data and link components together.

### Future ideas

- I think variable collision will likely be an issue. In the future, perhaps we can build features to address this. e.g. in the config builder showing when collisions happen and recommending signal aliasing, building locally scoped variables,  etc. I think @kmcq's FrameWrapper code could be an interesting location to do this aliasing as it is a shim that wraps all components.
- With global state all in one location, I think it should be easy for us to persist values in local storage or on the server. e.g. cursor location in a multi-stage workflow, cursor location in a workflow iterating on many results, etc..


kmcq (commented):
I have some bigger ideas about this, but gonna let them bake for a bit before I respond so for now you just get some simple feedback :)

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -19,14 +20,19 @@ interface FrameContext {
...
+  const processedProps = processProps({
+    viewProps,
+    COMPONENTS,
+    componentName,
+    signals,
+  })
   if (frames === undefined) {
     if (data) {
-      return <DataProvider {...processedProps} component={Component} />
+      return <DataProvider {...processedProps} data component={Component} />
```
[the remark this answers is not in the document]
hktouw:
```suggestion
      return (
        <DataProvider {...processedProps} data={data} component={Component} />
      )
```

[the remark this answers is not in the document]
hktouw: Good catch! It was already implicitly depending on data that was in `processedProps`. Since I removed that, I added it back in here.

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -73,14 +80,48 @@ function processProps({
...
+      const result = signalsType.safeParse(signal)
+      if (result.success) {
+        newProps[propName] = signal
+      }
+    }
+  })
+
   return {
-    data,
-    config,
```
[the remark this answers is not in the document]
hktouw: Mostly because it wasn't being used to my knowledge, and I _think_ config is not particularly useful for components in general as it is just yaml. We'd likely want to create other helpers and pass those in to the components instead. What do you think?

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -22,13 +28,19 @@ export function RenderConfig({
 }): React.JSX.Element {
   _COMPONENTS = COMPONENTS
   // Runs validations in config-schemas
+  console.log(COMPONENTS)
```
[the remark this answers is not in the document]
hktouw:
```suggestion
```

demo-app/src/open-truss/configs/6-state-management-demo.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - frame:
+      view:
+        component: IncrementAccountID
+        props:
+          accountIds: :userIds
+    - frame:
+      view:
+        component: IncrementAccountID
+        props:
+          accountIds: :userIds
```
hktouw, line 22: @kmcq shows alias feature that we talked about. Can see in this demo the two different sets of components that are reading from different signals

https://github.com/open-truss/open-truss/assets/4481584/04da64c3-1078-4380-9981-3967910fbea0

demo-app/src/open-truss/signals.ts
```diff
@@ -0,0 +1,12 @@
+import { SignalType } from '@open-truss/open-truss'
+
+// scalar types
+export const NumbersSignal = SignalType<number[]>('number[]', [])
```
hktouw, line 4: Added a handful of basic scalar signal types. We'll need to add more here if we want to use more scalars.

Using the `NumbersSignal` as an example, components wanting to use `number[]` for a prop type will need to import this Signal and add it to their prop zod type. e.g.

```typescript
  accountIds: NumbersSignal,
})
```

You would need to declare a corresponding signal in the config

```yaml
workflow:
  signals:
    accountIds: number[]
```

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -48,7 +52,14 @@ export function RenderConfig({
     configPath,
     COMPONENTS,
   ) as FrameWrapper
+
+  const signals = createSignals(config.signals)
```
hktouw, line 56: Updated the code to create signals based on what signals are declared in the config

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -48,7 +52,14 @@ export function RenderConfig({
...
+  const propErrors = validateComponentProps(config.frames, COMPONENTS, signals)
```
hktouw, line 57: Made a validation function for component props. Right now this is console logging any errors. We don't want to throw any errors so it doesn't break the config builder, which has a valid use case for "invalid component props".

packages/open-truss/src/signals/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  name: string,
+  defaultValue: T,
+  validator?: () => boolean,
+): z.ZodDefault<z.ZodType<Signal<T>>> {
+  validator ||= () => true
+
+  const zodType = z
+    .custom<Signal<T>>(validator)
+    .default((): Signal<T> => signal<T>(defaultValue))
+    .describe(`Signal<${name}>`)
```
hktouw, line 31: I just put `Signal<type>` as the description. If we need to hang anymore metadata on here in the future, let's turn this into a json object and create methods to parse metadata from that.

hktouw (commented):
@kmcq @jonmagic This is ready for another review at your convenience.

demo-app/src/open-truss/components/IncrementAccountID.tsx
```diff
@@ -0,0 +1,35 @@
...
+} from '@open-truss/open-truss'
+import { AccountIDsSignal } from '../signals'
+import { type z } from 'zod'
+import { useState } from 'react'
+
+export const Props = BaseOpenTrussComponentV1PropsShape.extend({
+  accountIds: AccountIDsSignal,
+})
+
+export default function IncrementAccountID({
```
kmcq, line 13: small nit: we're not incrementing anything here so maybe a better name is `CollectAccountID`?

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -126,20 +141,96 @@ function processProps({
...
+    throw new Error(`Signal passed to component is incorrect type.
+      Expected ${propName} to be type ${signalsType.description},
+      but got ${typeof signal}.
+    `)
+  }
+}
+
+function isSymbol(possibleSignal: any): boolean {
+  if (typeof possibleSignal !== 'string') return false
+  if (possibleSignal[0] !== ':') return false
```
kmcq, line 192: Could we use `String#startsWith`?

```suggestion
  if (!possibleSignal.startsWith(':')) return false
```

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -163,3 +254,22 @@ export function getComponent(
 
   return Component
 }
+
+export function getComponentProps(
```
kmcq, line 258: I feel like we could combine `getComponent` and `getComponentProps`. One idea is to have `getComponent` return the component and `Props` if it exists and then we could pass `Props` into `processProps` up above.

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -126,20 +141,96 @@ function processProps({
...
+  if (!props) return
+
+  Object.entries(props).forEach(([propName, propValue]) => {
+    const description = propValue?._def?.description
+    if (description) {
+      const signalsType = getSignalsType(description)
+      if (signalsType !== undefined) {
+        func(propName, signalsType)
+      }
+    }
```
kmcq, line 224: I think this code reads as if we're saying that if there's a description, it's a signal. We're not doing that though, so perhaps `getSignalsType` should just handle the case where there is no description and we replace this with something like this:

```suggestion
    const signalsType = getSignalsType(propValue)
    if (signalsType !== undefined) {
      func(propName, signalsType)
    }
```

I like that this moves the digging into zod internals into `getSignalsType` too. What do you think?

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
kmcq: I really like this error validation check! If we move this into `<Frame />`'s new `try ... catch` then it would be totally fine to throw errors as that code catches them and renders them to the user. What do you think about doing that?

packages/open-truss/src/signals/index.ts
```diff
@@ -0,0 +1,36 @@
+import { z } from 'zod'
+import { signal, type Signal } from '@preact/signals-react'
+export { effect, computed } from '@preact/signals-react'
+export type SignalsZodType = z.ZodDefault<z.ZodType<Signal<any>>>
+export type Signals = Record<string, Signal<any>>
+export type SignalTypes = Record<string, SignalsZodType>
+export const SIGNALS: SignalTypes = {}
```
kmcq, line 7: Might be nice to have a comment here about how this is used (I just had to look it up and it's not the easiest grep).

> We keep a global store of all SignalTypes to map the string of a signal's declared type e.g. `number[]` to the actual preact signal of that type.

kmcq (commented):
I haven't finished reading yet but here are some initial comments. This is so rad!

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
hktouw: I _think_ we want both? Soon when we'll be rendering frames selectively for multi-stage workflows, this creates a scenario where code paths for sub frames won't be exercised until the workflow is run. In that world, we want this validation where it is since it checks every frame regardless of if it is rendered or not, allowing us to confirm the validity of a workflow before it is run.

That being said, we probably also want validation at the Frame level as well since that will likely have more "runtime" data and we can perform validations we can't otherwise at config "compile" time.

packages/open-truss/src/signals/index.ts
hktouw: Good idea!

demo-app/src/open-truss/components/IncrementAccountID.tsx
hktouw: lol you're totally right. I think in my mind it started with just incrementing, but I wanted to use ya'll's ids in the demo, so I changed it.

hktouw:
@kmcq I implemented your feedback on this PR. Thanks for the review! This should be ready for another 👀

demo-app/src/open-truss/components/CollectAccountIDs.tsx
```diff
@@ -0,0 +1,35 @@
...
+} from '@open-truss/open-truss'
+import { AccountIDsSignal } from '../signals'
+import { type z } from 'zod'
+import { useState } from 'react'
+
+export const Props = BaseOpenTrussComponentV1PropsShape.extend({
+  accountIds: AccountIDsSignal,
+})
+
+export default function CollectAccountIDs({
```
kmcq, line 13: I know I just had you change the name, but maybe we should change it again 😸 As I'm reading this PR I'm realizing we should lean more into the demo app being documented examples. I'm picking on you a bit here because this is our first actually complicated component. That said, what do you think about naming this something like `SimpleSignalExample` and adding a comment here explaining what's going on?

>This component shows how Open Truss components can use signals (backed by @preact/signals-react) to share state with other components. [more here about how to use signals]

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
kmcq: Having it in both places makes sense. Could you move it into the `if (validateConfig)` block I added recently (a few dozen up)? I'm using that in the config builder for when we are okay with a config being wrong

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -126,20 +141,83 @@ function processProps({
...
+): void {
+  const result = signalsType.safeParse(signal)
+  if (result.success) {
+    newProps[propName] = signal
+  } else {
+    throw new Error(`Signal passed to component is incorrect type.
+      Expected ${propName} to be type ${signalsType.description},
+      but got ${typeof signal}.
+    `)
+  }
```
kmcq, line 178: This is really rad, thanks for doing the renaming! I think this deserves a comment saying what this unlocks either here or up where `addSignalToProps` is used.

demo-app/src/open-truss/configs/6-state-management-demo.yaml
kmcq: So rad! Let's for sure document this ability in the config spec

kmcq (approved):
So cool 🤩

PR #111, opened 2024-02-13
Config spec v0.0.2 proposal

hktouw:
## Why

To support more sophisticated workflows that need multiple stages, stored queries, and state management we should update the config spec to support these features.

## How

Please read the PR for a proposal for how to support these features. 

## Feedback requested

- General feedback
- Does this approach make sense? Is there a better way to accomplish this?
- What are considerations that are not being addressed here? Does this paint us into a corner in undesirable ways?

### DIFF
--- a/config-spec.md
+++ b/config-spec.md
@@ -104,7 +104,7 @@ workflow:
 // 3. Shows list of issues belonging to that account_id
 workflow:
   version: 1
-  signal:
+  signals:
     - account_id: number
   frameRender: sequence
   frames:
--- a/config-spec.md
+++ b/config-spec.md
@@ -106,7 +106,7 @@ workflow:
   version: 1
   signals:
     - account_id: number
-  frameRender: sequence
+  renderFrames: InSequence
   frames:
     - frame:
       view:
@@ -122,7 +122,7 @@ workflow:
           WHERE author_id = :account_id
         mapColumns:
           issue_id: issue_ids # TableOfIssues takes issue_ids
-        type: storedQuery
+        type: StoredQuery
         pageSize: 50
         source: mysql1
       frames:
--- a/config-spec.md
+++ b/config-spec.md
@@ -120,8 +120,11 @@ workflow:
           SELECT id AS issue_id
           FROM issues
           WHERE author_id = :account_id
-        mapColumns:
-          issue_id: issue_ids # TableOfIssues takes issue_ids
+        params:
+          account_id: :account_id
+        transforms:
+          mapColumns:
+            issue_id: issue_ids # TableOfContent
         type: StoredQuery
         pageSize: 50
         source: mysql1
@@ -131,7 +134,9 @@ workflow:
             component: OTStoredQueryToolbar
         - frame:
           view:
-            component: TableOfIssues
+            component: TableOfContent
+              content_ids: :issue_ids # example of argument aliasing
+              content_type: issue
 ```
 
 </details>
--- a/config-spec.md
+++ b/config-spec.md
@@ -97,11 +97,11 @@ workflow:
 ## Sample configuration
 
 ```yaml
-// Worfklow name: Get all issues for a given user
-// This is a multi-stage workflow
-// 1. Show input form to add account_id
-// 2. User inputs account_id
-// 3. Shows list of issues belonging to that account_id
+# Worfklow name: Get all issues for a given user
+# This is a multi-stage workflow
+# 1. Show input form to add account_id
+# 2. User inputs account_id
+# 3. Shows list of issues belonging to that account_id
 workflow:
   version: 1
   signals:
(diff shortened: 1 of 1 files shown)


config-spec.md
```diff
@@ -65,3 +67,71 @@ workflow:
...
+
+```yaml
+// Worfklow name: Get all issues for a given user
+// This is a multi-stage workflow
+// 1. Show input form to add account_id
+// 2. User inputs account_id
+// 3. Shows list of issues belonging to that account_id
+workflow:
+  version: 1
+  signal:
```
kmcq, line 107:
```suggestion
  signals:
```

kmcq (commented):
Really interesting stuff! Question about the versioning: as far as I can tell this includes no breaking changes to the existing spec. Do you see this causing us to version the engine? Right now it doesn't really matter, but I'd love to limit versioning the engine as much as possible as it might be annoying

hktouw:
> Question about the versioning: as far as I can tell this includes no breaking changes to the existing spec. Do you see this causing us to version the engine? Right now it doesn't really matter, but I'd love to limit versioning the engine as much as possible as it might be annoying

I don't think anything in here should be a breaking change! And fwiw, I am of the opinion that it is probably fine to do breaking changes until we get to a V1 for the reasons you mentioned.

config-spec.md
```diff
@@ -65,3 +67,71 @@ workflow:
...
+            - :account_id
+    - frame:
+      data:
+        query: >\n
+          SELECT id AS issue_id
+          FROM issues
+          WHERE author_id = :account_id
+        mapColumns:
+          issue_id: issue_ids # TableOfIssues takes issue_ids
+        type: storedQuery
```
jonmagic, line 125: Not sure why but feeling like this should be capitalized.

```suggestion
        type: StoredQuery
```

config-spec.md
```diff
@@ -65,3 +67,71 @@ workflow:
...
+# Worfklow name: Get all issues for a given user
+# This is a multi-stage workflow
+# 1. Show input form to add account_id
+# 2. User inputs account_id
+# 3. Shows list of issues belonging to that account_id
+workflow:
+  version: 1
+  signals:
+    - account_id: number
+  frameRender: sequence
```
jonmagic, line 109: Can the value of this be a "component", i.e. a pascal cased function?

```suggestion
  renderFrames: InSequence
```

config-spec.md
```diff
@@ -65,3 +67,71 @@ workflow:
...
+        props:
+          fields:
+            - :account_id
+    - frame:
+      data:
+        query: >\n
+          SELECT id AS issue_id
+          FROM issues
+          WHERE author_id = :account_id
+        mapColumns:
```
jonmagic, line 123: We talked about changing this to be a `transforms` key that can take in a list of map reduce functions I think?

jonmagic (approved):
All of my suggestions are merely suggestions. I'm going to approve so we can keep moving forward quickly.

config-spec.md
hktouw: Perhaps? Maybe we can make this such that we give the function / component the config, cursor, and other relevant vars and it returns what frames to render? 

We'll have to play around with that though and see if that makes sense. I'm 👍 with going with your change for now and we can update the annotation to something else if that doesn't end up working.

config-spec.md
hktouw: Sounds good to me 👍

config-spec.md
```diff
@@ -65,3 +67,76 @@ workflow:
...
+        pageSize: 50
+        source: mysql1
+      frames:
+        - frame:
+          view:
+            component: OTStoredQueryToolbar
+        - frame:
+          view:
+            component: TableOfContent
+              content_ids: :issue_ids # example of argument aliasing
```
hktouw, line 138: @kmcq here is an example of your aliasing idea that we can do in processProps.

config-spec.md
```diff
@@ -65,3 +67,76 @@ workflow:
...
+        props:
+          fields:
+            - :account_id
+    - frame:
+      data:
+        query: >\n
+          SELECT id AS issue_id
+          FROM issues
+          WHERE author_id = :account_id
+        params:
```
hktouw, line 123: Per our conversation, we're doing to just declare parameters here as opposed to trying to extract them from the query. Furthermore, we don't have any type information in the query and couldn't do this anyway.

Pie in the sky idea 💡. In the distant future when we support integration with data catalogs, I wonder if we could get type information from that and apply it to the query? I feel like having queries dynamically typed would be a great feature. Another way we could do this too perhaps would be to run explain queries and extract type information that way.

hktouw:
Going to merge this for now since this is a great start and we have something to work with. I'll open up issues for the various upgrades to the rendering engine that need to be made to support this config spec.

PR #113, opened 2024-02-16
Refactor types to make them more descriptive

kmcq:
This PR does two main things:

1. `withChildren` is just an object now. The function way was throwing away types and after refactoring it to not do this I realized that I think this is more ergonomic.
2. `BaseOpenTrussComponentV1` is now a generic type that can be used to declare components. I think this is nice as we don't need to type every component as returning JSX.Element.

**Before**
image

**After**
image

### DIFF
--- a/demo-app/src/open-truss/components/TrinoDemo.tsx
+++ b/demo-app/src/open-truss/components/TrinoDemo.tsx
@@ -1,16 +1,20 @@
 'use client'
 import {
+  type BaseOpenTrussComponentV1,
   BaseOpenTrussComponentV1PropsShape,
   withChildren,
 } from '@open-truss/open-truss'
 import { z } from 'zod'
 
-export const Props = withChildren(BaseOpenTrussComponentV1PropsShape).extend({
+export const Props = BaseOpenTrussComponentV1PropsShape.extend({
+  ...withChildren,
   data: z.string().default('no data'),
 })
 
-export default function TrinoDemo({
+const TrinoDemo: BaseOpenTrussComponentV1<z.infer<typeof Props>> = ({
   data,
-}: z.infer<typeof Props>): JSX.Element {
+}) => {
   return <>{JSON.stringify(data)}</>
 }
+
+export default TrinoDemo
--- a/packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
+++ b/packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
@@ -95,32 +95,33 @@ export type WorkflowV1 = z.infer<typeof WorkflowV1Shape>
 
 export const BaseOpenTrussComponentV1PropsShape = z.object({
   data: DataV1Shape,
-  config: WorkflowV1Shape,
+  config: WorkflowV1Shape.optional(),
 })
-export const withChildren = (shape: z.AnyZodObject): z.AnyZodObject =>
-  shape.extend({ children: z.any().optional() })
-const ComponentWithChildrenShape = withChildren(
-  BaseOpenTrussComponentV1PropsShape,
-)
+
+export const withChildren = { children: z.any().optional() }
+const ComponentWithChildrenShape =
+  BaseOpenTrussComponentV1PropsShape.extend(withChildren)
 type ComponentWithChildren = z.infer<typeof ComponentWithChildrenShape>
 
-export type BaseOpenTrussComponentV1Props =
-  | z.infer<typeof BaseOpenTrussComponentV1PropsShape>
-  | ComponentWithChildren
-
-export type BaseOpenTrussComponentV1 = (
-  props: BaseOpenTrussComponentV1Props,
-) => React.JSX.Element
-
-const FrameWrapperShape = withChildren(
-  BaseOpenTrussComponentV1PropsShape,
-).extend({
+const FrameWrapperShape = BaseOpenTrussComponentV1PropsShape.extend({
+  ...withChildren,
   configPath: z.string(),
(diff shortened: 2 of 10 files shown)


jonmagic (approved):
I love it 👍 Great work. For reference while reviewing I had a _chat with gpt_ about the differences between function declaration and arrow function expression to make sure the changes were primarily for ergonomics: https://chat.openai.com/share/ed5782e6-2008-4a16-93fe-762d807b4156

demo-app/src/open-truss/components/TrinoDemo.tsx
```diff
@@ -1,16 +1,20 @@
 'use client'
 import {
+  type BaseOpenTrussComponentV1,
   BaseOpenTrussComponentV1PropsShape,
   withChildren,
```
hktouw, line 5: Thinking out loud here, what do we think about having a `BaseOpenTrussComponentV1WithChildrenPropsShape`? Children will probably be used a lot, so having this saves a couple lines (import + use). Pretty chatty name of course..

packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
```diff
@@ -95,32 +95,33 @@ export type WorkflowV1 = z.infer<typeof WorkflowV1Shape>
...
 
-const FrameWrapperShape = withChildren(
-  BaseOpenTrussComponentV1PropsShape,
-).extend({
+export const withChildren = { children: z.any().optional() }
+const ComponentWithChildrenShape =
+  BaseOpenTrussComponentV1PropsShape.extend(withChildren)
+type ComponentWithChildren = z.infer<typeof ComponentWithChildrenShape>
+
+const FrameWrapperShape = BaseOpenTrussComponentV1PropsShape.extend({
```
hktouw, line 116: `BaseOpenTrussComponentV1PropsShape`'s properties are already included when you use the `BaseOpenTrussComponentV1<>` generic on line 122. Do you need to extend this shape or can you just use `z.object`?

demo-app/src/open-truss/components/TrinoDemo.tsx
```diff
@@ -1,16 +1,20 @@
...
 } from '@open-truss/open-truss'
 import { z } from 'zod'
 
-export const Props = withChildren(BaseOpenTrussComponentV1PropsShape).extend({
+export const Props = BaseOpenTrussComponentV1PropsShape.extend({
+  ...withChildren,
   data: z.string().default('no data'),
```
hktouw, line 11: This is an interesting prop. `data` is already declared in the `BaseOpenTrussComponentV1Props` type as `YamlType | undefined`. Redeclaring it here effectively extends it to also have `string` as a valid type, hence `(YamlType | undefined) & string` in Keith's screenshot.

I _think_ we may want to go the route of removing data from the base type altogether and not pass in the entire data object to every component. This forces components to add an interface for the data they need and will make composability easier as we have an interface to work off of. Does that seem like a reasonable path to go down? 

Of course we could have a handful of components that legitimately take in the entire data payload, but that would be less common I think. It is a case where the expectation is the component would have a "Data" higher order type as an interface and the entire data payload makes sense for those use cases (e.g. rendering the whole blob to the page for debugging, saving an entire blob off to storage for logging, etc..).

hktouw (commented):
Great work Keith! Also, I feel like I've leveled up my typescript game a bit by reading this PR 🤘.

demo-app/src/open-truss/components/TrinoDemo.tsx
kmcq: I considered that! I thought it was too wordy hah but I see your point about number of lines... Should we drop the `BaseOpenTruss*` prefix?

- `ComponentV1`
- `ComponentWithChildrenV1Shape`

demo-app/src/open-truss/components/TrinoDemo.tsx
kmcq: Good catch @hktouw -- I thought this was just a dummy prop that @jonmagic added at some point and that it was incorrect (and that it was okay because this component's not done yet). I didn't realize that zod's `.extend` adds a type union. This is a good nudge to consider what you're saying 👍 

> I think we may want to go the route of removing data from the base type altogether and not pass in the entire data object to every component.

After seeing how `.extend` works, I agree. I think we could help with this and add something akin to but more advanced than the `withChildren` I created in this PR that is a function that helps folks create a `data` type that adhere's to what `uqi` will fetch and is typed correctly for the type of query results the component expects.

```ts
  data: uqiResultShape({ rowShape: z.object({ accountId: z.number(), login: z.string() }) }),
})

// then, in the component

props.row.map(({ accountId, login }) => {
  // `accountId` has type number
  // `login` has type string
})
```

Pretty sure this is possible to do 😅

packages/open-truss/src/configuration/engine-v1/config-schemas.tsx
kmcq: You're correct, this is duplicative. I did this because it's how OT-compatible components do it and `FrameWrapper` is one so I did the same thing. I hadn't thought of just doing `object` but I think it would be equivalent. I think I like this way better because it keeps things aligned and we do want regular OT components to do this by extending the base shape (we `export const Props` those ones and if it were just an extension of `object` it would miss the base props). What do you think?

hktouw: That makes sense to me!

hktouw (commented):
🥳

PR #114, opened 2024-02-16
Use shadcn/ui for components, themes, and styling in the demo-app

jonmagic:
## Why?

We haven't started using any kind of component or styling libraries for the demo-app yet.

## How?

I propose we use shadcn/ui which is radix and tailwindcss under the hood.

### Before

CleanShot 2024-02-16 at 13 47 42@2x

### After

CleanShot 2024-02-16 at 13 46 47@2x

## Considerations

We won't necessarily be using this component library or tailwind for the work projects that will use open-truss but it felt like a good option for the demo-app.

TBD how we'll choose when to build an open-truss component in the demo-app vs the `@open-truss` packages but I think this will become more obvious as we continue to build out the demo app and then start using open-truss in TDJ (the day jobs).


kmcq:
Awesome stuff! It made the config builder quite difficult to use. What should we do about that? Add some default button and input styles?

image

jonmagic:
> What should we do about that? Add some default button and input styles?

Great question. I think this is what forces us to decide when component is purely functional and when it is about display. The functional ones should go in the open-truss package and the visual ones (buttons, input, textarea, etc) should go in the demo-app I think.

I wonder if this means a lot of the open-truss package will be react hooks vs UI components 🤔

hktouw:
> I wonder if this means a lot of the open-truss package will be react hooks vs UI components

For OT pages, how do we feel about being more opinionated about that and picking the UI / styles we want to? This would be pages like the config builder, workflow sessions, playground, etc but any actual workflows will use whatever styles adopters have made their components with.

In terms of UI development of OT components in the OT package, that may be a bit tricker. My guess is where we can we can refactor functionality into their own components and add that to the base package. For UI components, we can also go the route of creating specialized OT component packages per framework. e.g. `open-truss/primer` or `open-truss/primer-components`. That may strike the best balance of making it easy to open source UI components, while not making development painful having to account for multiple different frameworks. 

Additionally, I bet we could make AI tools / workflows that relatively easily converts components to other UI frameworks with high fidelity. I bet this workflow can even be an Open Truss Workflow 😉

jonmagic:
CleanShot 2024-02-26 at 14 49 13@2x

jonmagic:
Note this doesn't convert the available workflows UI yet because that is still inside of the open-truss package.

hktouw (approved):
LGTM!

PR #115, opened 2024-02-16
workflows are frames; render them like they are

kmcq:
Consolidates rendering logic into `<Frame />`

### DIFF
--- a/packages/open-truss/src/configuration/engine-v1/Frame.tsx
+++ b/packages/open-truss/src/configuration/engine-v1/Frame.tsx
@@ -52,6 +52,23 @@ export function Frame(props: FrameContext): React.JSX.Element {
   } = props
   const { component, props: viewProps } = view
   try {
+    const subframes = frames?.map((subframe, k) => {
+      const subframePath = `${configPath}.frames.${k}`
+      return (
+        <FrameWrapper key={k} frame={subframe} configPath={subframePath}>
+          <Frame
+            frame={subframe}
+            configPath={subframePath}
+            globalContext={props.globalContext}
+          />
+        </FrameWrapper>
+      )
+    })
+
+    if (component === '__FRAGMENT__') {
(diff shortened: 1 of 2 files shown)


packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -41,31 +42,28 @@ export function RenderConfig({
...
   ) as FrameWrapper
   const globalContext: GlobalContext = {
     config,
     COMPONENTS,
     FrameWrapper,
   }
+  // Workflows are just frames! Use unknown to convince TS of this fact.
+  const frame = {
+    ...config,
+    view: { component: '__FRAGMENT__' },
```
kmcq, line 59: Not sure about this hack yet. Need something to say this is top level and it's okay to not have a component

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -52,6 +52,23 @@ export function Frame(props: FrameContext): React.JSX.Element {
...
+          <Frame
+            frame={subframe}
+            configPath={subframePath}
+            globalContext={props.globalContext}
+          />
+        </FrameWrapper>
+      )
+    })
+
+    if (component === '__FRAGMENT__') {
```
kmcq, line 68: See below comment about how I'm not sure about this

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
jonmagic: Hacks FTW

jonmagic (approved):
I'm cool with the hack. We can always roll back 😄

hktouw (commented):
Love this! Great to see us consolidate logic into Frame 🤘

PR #118, opened 2024-02-22
Move config builder's UI to the demo app

kmcq:
We're working to move display components out of the main package and into the demo app; this PR handles the config builder.

To do this I've:

- Moved the config builder context to a new `hooks/` directory and updated what it exports
- Moved the UI to the demo app
  - This involved some changes to account for where things are imported, etc

I haven't done any work to make this look good yet. I was thinking we'd merge this PR, get it into #114, and then update that PR to make this look nice.


packages/open-truss/src/hooks/config-builder.tsx
```diff
@@ -1,37 +1,48 @@
...
   setConfig: (config: string) => void
   framesPath: string
-  setFramesPath: (path: string) => void
+  setFramesPath: (path: string | null) => void
   addFrame: (frame: FrameType) => void
   deleteFrame: (framePath?: string) => void
 }
-export const ConfigBuilderContext = createContext<ConfigBuilder>({
+
+const ConfigBuilderContext = createContext<ConfigBuilder>({
```
kmcq, line 24: FYI I've learned that exporting the result of `createContext` from a package doesn't work well, something to do with types getting weird. Instead we should export a hook (e.g. `useConfigBuilderContext` below) that uses the context.

jonmagic (approved):
Super cool! There is some sort of infinite render loop bug, I'll see if I can figure it out:

CleanShot 2024-02-24 at 09 08 06

EDIT: Turns out this bug exists on main too 😄

jonmagic:
> There is some sort of infinite render loop bug

It comes from https://github.com/open-truss/open-truss/blob/7b1e1d1c3cab6fd7d1c57b5f27d92032f0cf900d/packages/open-truss/src/pages/config-builder-page/PropInput.tsx#L19-L21

jonmagic:
> After this, the goal is to move display components into framework specific packages right? e.g. we'll move the config builder display components to `open-truss/shadcn`.

At some point probably 😄 I'm in no rush to do that step though.

Issue #112, opened 2024-02-16
[Milestone] Config spec v0.0.2


hktouw:
## Why

To support more complex workflows (e.g. multi-stage workflows, stored query backend, pagination through results), we need a more sophisticated version of the config spec.

## How

We will to implement the next version of the config spec as proposed in https://github.com/open-truss/open-truss/pull/111.

<details>

<summary>Example configuration</summary>

```yaml
# Worfklow name: Get all issues for a given user
# This is a multi-stage workflow
# 1. Show input form to add account_id
# 2. User inputs account_id
# 3. Shows list of issues belonging to that account_id
workflow:
  version: 1
  signals:
    - account_id: number
  renderFrames: InSequence
  frames:
    - frame:
      view:
        component: OTInputForm
        props:
          fields:
            - :account_id
    - frame:
      data:
        query: >\n
          SELECT id AS issue_id
          FROM issues
          WHERE author_id = :account_id
        params:
          account_id: :account_id
        transforms:
          mapColumns:
            issue_id: issue_ids # TableOfContent
        type: StoredQuery
        pageSize: 50
        source: mysql1
      frames:
        - frame:
          view:
            component: OTStoredQueryToolbar
        - frame:
          view:
            component: TableOfContent
              content_ids: :issue_ids # example of argument aliasing
              content_type: issue
```

</details>

### Tasks
- [x] Decide on first pass of config spec v0.0.2 https://github.com/open-truss/open-truss/pull/111
- [ ] Implement state management https://github.com/open-truss/open-truss/pull/110
- [ ] Implement renderFrames
- [ ] Implement navigation functions
- [ ] Implement data params
- [ ] Implement data transforms
- [ ] Implement StoredQueries

PR #119, opened 2024-02-23
Support renderFrames annotation

hktouw:
ref https://github.com/open-truss/open-truss/issues/112

### Why

Per config spec v0.0.2, we want to develop a `renderFrames` annotation that can be used to move between frames and sequence rendering. This supports multi-stage workflows.

### How

Using the following config:

```yaml

workflow:
  version: 1
  id: 7-render-frames-demo-unique-id
  signals:
    integer: number
    runCalculation: NavigateFrame
  renderFrames:
    type: inSequence
    next: :runCalculation # sets the next function as :runCalculation
  frames:
    - frame:
      view:
        component: IntegerInput
        props:
          submit: :runCalculation # sets submit to runCalcluation
    - frame:
      view:
        component: ShowIntegerSquared
```

We are able to generate a multi-stage workflow where we first show an input field, accept user input, then render the result.

https://github.com/open-truss/open-truss/assets/4481584/97c6a0c9-6a16-4882-b116-569de2ffb3dd

Interestingly, we are able to change the render type back to the default and re-use the components as expected. This shows how components can be re-usable for different workflows / use cases.

```yaml
workflow:
  version: 1
  id: 7-render-frames-demo-unique-id
  signals:
    integer: number
  frames:
    - frame:
      view:
        component: IntegerInput
    - frame:
      view:
        component: ShowIntegerSquared

```

https://github.com/open-truss/open-truss/assets/4481584/438a1d11-7b6a-4f48-861c-2cdf79d7596a


demo-app/src/open-truss/configs/7-render-frames-demo.yaml
```diff
@@ -0,0 +1,62 @@
...
+# x write out config
+# x build display components
+# x create renderFrames functionality in engine
+# x create default renderFrames function (All)
+# x create InSequence render function
+#   x render selected frame
+#   x set next function in signal store
+#   x create nav signal type
+# x [skip] create navigation? prop type
+# x [skip] update processProp code to pass in navigation functions to components
```
jonmagic, line 62:
```suggestion
```

demo-app/src/open-truss/configs/7-render-frames-demo.yaml
```diff
@@ -0,0 +1,62 @@
+workflow:
+  version: 1
+  id: 7-render-frames-demo-unique-id
+  signals:
+    integer: number
+    runCalculation: NavigateFrame
+  renderFrames:
+    type: inSequence
+    next: :runCalculation # sets the next function as :runCalculation
```
jonmagic, line 9: Been thinking about the naming of things a bit and wonder if shortening `renderFrames` to `render` and changing `type: inSequence` to `mode: sequential` reads a bit better 🤔 

```
  render:
    mode: sequential
    next: :runCalculation
```

The default mode which could be omitted:
```
  render:
    mode: parallel
```

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -123,3 +133,26 @@ function validateComponentProps(
...
+
+export function setWorkflowSessionValue(
+  id: string,
+  key: string,
+  value: string | number,
+): void {
+  const session = getWorkflowSession(id)
+  session[key] = value
+  localStorage.setItem(`OT-Workflow:${id}`, JSON.stringify(session))
+}
```
jonmagic, line 158: The store for this will need to be configurable right? Maybe add a TODO here or open an issue?

jonmagic (approved):
I think this is rad 😄 Left a few small notes, nothing jumps out to me as a blocker to giving this a try.

demo-app/src/open-truss/configs/7-render-frames-demo.yaml
hktouw: Interesting wrinkle is `renderFrames` explicitly applies only to how we render the sub frames if they exist. If we ever need an annotation that determines how to render the frame itself then we'd need something like `renderFrame`. 

Perhaps we can run with `renderFrames` for now and after a while if that seems sufficient we can rename to `render`?

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
hktouw: Yup definitely. There is very hacky right now and we'll need to buff this out a lot more. Also we'll need to figure out how we want this data to interact or not interact with the url.  

Added `Implement client / server storage mechanism` to the config spec v0.0.2 milestone issue https://github.com/open-truss/open-truss/issues/112

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -107,6 +102,102 @@ export function Frame(props: FrameContext): React.JSX.Element {
...
+    configPath,
+  } = props
+  if (renderFrames?.type !== 'inSequence') {
+    throw new FrameError(`This error should never occur`, component, configPath)
+  }
+
+  const frameLevel = `${configPath}.frames`
+  const frameToRender = getCurrentFrameCursor(workflowId, frameLevel)
+
+  const nextFuncName = parseSignalName(renderFrames?.next)
```
hktouw, line 137: This creates a `next` function that moves to the next frame. If we want to make a `back` function we would add something similar here.

demo-app/src/open-truss/configs/7-render-frames-demo.yaml
```diff
@@ -0,0 +1,19 @@
+workflow:
+  version: 1
+  id: 7-render-frames-demo-unique-id
```
hktouw, line 3: Worth noting that we use the workflow id to generate a workflow session object in local storage scoped by the id. This is a hack right now and we'll eventually want a better way for generating unique ids for workflows and workflow sessions.

kmcq (commented):
oh no wonder you didn't know about my comments, I didn't press submit!

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -58,23 +59,17 @@ function ShowError({ error }: { error: FrameError }): JSX.Element {
...
-          <Frame
-            frame={subframe}
-            configPath={subframePath}
-            globalContext={props.globalContext}
-          />
-        </FrameWrapper>
-      )
-    })
+    const [renderCount, render] = useState(0)
+    // used to re render current Frame and it's tree
```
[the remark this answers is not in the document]
kmcq: I'll fix this :) gonna do a bit of organization in this file

kmcq:
This is very exciting @hktouw! #126 has my ideas on this, mostly nitpicks

PR #120, opened 2024-02-24
Create open-truss/shadcn package

hktouw:
### Why

We are moving towards hosting display components on a per front-end framework basis as a separate package. We should setup our project to support this.

### How

This PR scaffolds an `open-truss/shadcn` package. As a demo, the package imports files from the `open-truss/open-truss` package and exposes a `helloWorld` function. The demo-app then imports the `helloWorld` function and executes it as you would expect.

‎localhost3000 2024-02-24 at 1 33 39 PM

### Notes

- @jonmagic @kmcq please kick the tires on this and see if this all seams reasonable
- I updated `script/server` to also start up the `open-truss/shadcn` package. It is worth noting since we are deleting then rebuilding packages on `script/server` you may notice some transient compilation errors. e.g. compiling `open-truss/shadcn` will complain because `open-truss/open-truss` was deleted, but then it should resolve itself.


jonmagic (commented):
Can we punt on this for a bit or do you think this needs to happen now?

hktouw:
Punting for a bit is probably fine

hktouw:
Closing this out for now and we can open a new package when we need to.

PR #121, opened 2024-02-26
Fix render bug in config builder

jonmagic:
1. Not really sure how to fix the infinite render loop so I just made this change for now: 783245204e99b140a00d5c8d5c8fc6fc27c01dc5
2. These imports should come from open-truss and not the demo-app: cca6a9f32fdfeae738423f8a35546753889b4fc3

Still one outstanding bug afaict related to signals:

CleanShot 2024-02-26 at 10 23 31@2x

App go boom 💥 

CleanShot 2024-02-26 at 10 25 34@2x

### DIFF
--- a/demo-app/src/pages/ot/config-builder/PropInput.tsx
+++ b/demo-app/src/pages/ot/config-builder/PropInput.tsx
@@ -1,6 +1,8 @@
+import {
+  type YamlType,
+  type ZodDescriptionObject,
+} from '@open-truss/open-truss'
 import { useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


hktouw (approved):
🥳

jonmagic:
I'm going to merge this but just want to note that 1 above may not be the correct fix (cc @kmcq) and that the error screenshots at the bottom of the OP are an ongoing issue. Those can be addressed in follow up PRs though. Shipping this PR will let me update the shadcn PR to apply styles to the config builder.

PR #126, opened 2024-02-29
A bit of cleanup for renderFrames

kmcq:
Not too much here, but I made some things into components and cleaned up how props are passed around


jonmagic (approved):
🤘

PR #128, opened 2024-03-01
First pass at data provider component

hktouw:
Paired with @jonmagic 

### Why

We need the ability to support flexibly querying data in OT

### How

This PR takes the approach of mostly re-using the signals architecture we built and building a Data Provider component, instead of developing a dedicated `data` annotation.

‎localhost:3000otplayground4-trino-demo 2024-02-29 at 7 55 30 PM

### Notes

- We moved the scalar signals to the OT package so that they can be used by the `OTUqiDataProvider` component
- We made a first pass at making the `OtUquiDataProvider` component
  - This component accepts a `query` and `source` argument to use for issuing a uqi query. The `output` argument specifies the signal name and the shape of the data that will be passed into the signal. If the shape is an array with an object, it will validate that the uqi rows have the same field and type names and return all the uqi results. If the shape is an object, it will do this check on just the first row and return just the first row.
  - This is an MVP data provider component that we want to ship so Jon can start using. There will be many follow up PRs necessary to improve this. See below.

### Follow ups

- Update component to only set the fields that the output has declared. Currently it just validates the fields exist in the results, but sets all the fields into the signal
- Support pagination
- Update component to set `isLoading` signal that can be used by display components to render loading indicators. Add `isLoading` to the base open truss component type so it's available in all components
- Update signals so that you can declare them at every frame level and signal name resolution moves up recursively up the frame parent tree. Similar to variable scoping in programming and how you can access variables in outer closures.
- Update component to support output annotations that are scalars instead of just arrays of objects or objects

### DIFF
--- a/demo-app/src/open-truss/configs/4-trino-demo.yaml
+++ b/demo-app/src/open-truss/configs/4-trino-demo.yaml
@@ -1,9 +1,113 @@
 workflow:
   version: 1
+  signals:
+    results: Records
   frames:
     - frame:
-      data:
-        query: select * from  tpch.sf1.customer limit 10
-        source: trino-demo
+      # signals:
+      #   results: Records
       view:
-        component: TrinoDemo
+        component: UqiDataProvider
+        props:
+          source: trino-demo
+          query: >
+            SELECT Column
+            FROM tpch.sf1.customer
+            limit 10
+          results: :results
+        # TODO modify workflow where you select the type
+        # and it gets set as a param here
+        # params:
+        #   type:
+        # pageSize: 50
+          output:
+            results:
+              - column: string
+                # Type: string
+                # Extra: string
+                # Comment: string
+            # results2:
+            #   key: string
+            #   type: string
+            #   value: string
+      frames:
+        - frame:
+          view:
+            component: TrinoDemo
+            props:
+              results: :results # TrinoDemo takes a colleciton type
+
+workflow:
+  version: 1
+  signals:
+    results: Records
+  frames:
+    - frame:
+      view:
+        component: UqiDataProvider
+        props:
+          source: trino-demo
+          query: >
+            SELECT Column
+            FROM tpch.sf1.customer
+            limit 10
+          results: :results
+          output:
+            results:
+              - Column: string
+                Type: string
+                Extra: string
+                Comment: string
+      frames:
+        - frame:
+          view:
+            component: TrinoDemo
+            props:
+              results: :results
+
+
+
+
+
+
+
+
+
+
+
+
+# workflow:
+#   version: 1
+#   signals:
+#     results: Records
+#   frames:
+#     - frame:
+#       data:
+#         query: >\n
+#           SELECT key, type, value
+#           FROM tpch.sf1.customer
+#           WHERE type = 'BigInt'
+#         source: trino-demo
+#         output:
+#           results:
+#             - key: string
+#               type: string
+#               value: string
+#       view:
+#         component: TrinoDemo
+#         props:
+#           results: :results # TrinoDemo takes a colleciton type
+# TODO
+# - [x] update trino demo component to take in results
+# - [ ] update data provider to support outputs
+#   - [x] sketched out config
+#   - [ ] update data provider to loop through output config and create signals for each property
+#   - [ ] grab data from the results and put them into the signals
+# - [ ] update data provider to support params (sequence demo_
+# - [ ] update data provider to support pagination
+# - [ ] update data provider to support isLoading
+# - [ ] update signals to have one at every frame level
+# - [ ] refactor component to OT package
+# - [ ] update UqiDataprovider type mapping to support user defined higher order signal types
+# - [ ] update UqiDataprovider to support annotgations that are scalar results and not just arrays of objects or objects
+# - [ ] update UqiDataprovider to only pass through results that the config has declared and not excess results returned by uqi.
--- a/demo-app/src/open-truss/configs/4-trino-demo.yaml
+++ b/demo-app/src/open-truss/configs/4-trino-demo.yaml
@@ -5,7 +5,7 @@ workflow:
   frames:
     - frame:
       view:
-        component: UqiDataProvider
+        component: OTUqiDataProvider
         props:
           source: trino-demo
           query: >
--- a/demo-app/src/open-truss/configs/4-trino-demo.yaml
+++ b/demo-app/src/open-truss/configs/4-trino-demo.yaml
@@ -4,51 +4,12 @@ workflow:
     results: Records
   frames:
     - frame:
-      # signals:
-      #   results: Records
       view:
         component: UqiDataProvider
         props:
           source: trino-demo
           query: >
-            SELECT Column
-            FROM tpch.sf1.customer
-            limit 10
-          results: :results
-        # TODO modify workflow where you select the type
-        # and it gets set as a param here
-        # params:
-        #   type:
-        # pageSize: 50
-          output:
-            results:
-              - column: string
-                # Type: string
-                # Extra: string
-                # Comment: string
-            # results2:
-            #   key: string
-            #   type: string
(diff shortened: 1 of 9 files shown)


demo-app/src/open-truss/configs/4-trino-demo.yaml
```diff
@@ -1,9 +1,27 @@
...
       view:
-        component: TrinoDemo
+        component: OTUqiDataProvider
+        props:
+          source: trino-demo
+          query: >
+            SELECT name, phone, acctbal, comment
+            FROM tpch.sf1.customer
+            limit 10
+          results: :results
```
hktouw, line 15: One thing that is a bit magical that I'm not the biggest fan of is this `results` prop is not in the components type definition. The component doesn't know what signals the user is going to set ahead of time so we couldn't model it here. Furthermore, what isn't obvious is the name of the key needs to match the name of `results` 2 lines down. For every key in `output` the component looks for a key on `prop` with that name and assumes it's a signal that it can throw the data into. 

This might be more verbose, but one thought is we could do an annotation like this

```yaml
output:
  results:
    signal: :results
    type:
      - name: string
      - phone: string
      - acctbal: number
      - comment: string
```

We would need to update the process props signals code to recursively go through the props and set anything that is a `:symbol` with a signal, but this way would remove the hidden dependency between output and other properties. Additionally, this will allow us to model the `output` prop better in the component's prop interface instead of looking for props that aren't declared by the interface.

demo-app/src/open-truss/configs/4-trino-demo.yaml
```diff
@@ -1,9 +1,27 @@
...
-      data:
-        query: select * from  tpch.sf1.customer limit 10
-        source: trino-demo
       view:
-        component: TrinoDemo
+        component: OTUqiDataProvider
+        props:
+          source: trino-demo
+          query: >
+            SELECT name, phone, acctbal, comment
```
hktouw, line 12: @jonmagic I was able to get this to work. The previous columns we were using (e.g. `Column`, `Type`, `Extra`) just didn't exist on this table at all and it wasn't a capitalization issue. Did you accidentally get those column names from somewhere else?

packages/open-truss/src/components/OTUqiDataProvider.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+// validates if the data conforms to the shape provided.
+// shape - key is the name of the property and the value is the javascript type
+// data - json object with followig shape
+//  [
+//    { key: 'greeting',  type: 'String', value: 'hello' },
+//    { key: 'goodbye',  type: 'String', value: 'byebye' }
+//  ]
+//
+function isShapeValid(
```
kmcq, line 99: Ooh I wonder if with #129 knowing about each signal's `valueShape`, we could just do a simple zod parse for this! Like `signal.valueShape.parse(queryResult.rows)` (or `.rows[0]`). This would require to have the top level `results: Record` definition be more specific, which is maybe a good thing? I wonder if we could have a way to declare signals completely from the config. Now that I'm getting this far, I realize this might be out of scope of this PR 😸

kmcq (commented):
Very cool! See my comment about pulling from the global store, I think we can make this just like you want it to be

demo-app/src/open-truss/configs/4-trino-demo.yaml
hktouw: @kmcq Jon and I were reviewing this and noticed that a consequence of doing this is workflows / components would not be declaring what signals they are using which would make it harder to connect related workflows to each other. Do you think that is the case as well? If not, are there other ways we can know how to connect workflows / components togethers?

kmcq: This is a great concern! I'm realizing that I don't agree with my last sentence. It's best if components declare their signals 👍

kmcq: That said, it does seem like a way for us to determine which components are compatible from reading configs might be really helpful. It would also let us have more structured signals without changing the code. For example let's say we had a query that returned a User and a User is a defined signal:


```ts
interface User {
  id: number
  name: string
}
const UserSignal = createSignal<User>(...)
```

```yaml
workflow:
  version: 1
  signals:
    results: UserSignal
  frames:
    - frame:
      view:
        component: OTUqiDataProvider
        props:
          source: trino-demo
          query: SELECT id, name FROM users
          results: :results
          output:
            results:
              - name: string
                id: number
```

If we wanted to also query for a user's email address or whatever, we'd have to either use a less defined signal (like the Records you've added in here) or also update the code definition of `UserSignal`.

I think my preferred way of doing this is what I said to you over tiki drinks the other day 🍹 What if we could declare the signal in the config?

```yaml
workflow:
  version: 1
  signals:
    results:
      - name: string
      - id: number
      - email: string
```

This idea is not fully baked :)

hktouw: Interesting! So the type definition is defined in the signal and the data component formats the results to conform with that? e.g.

```yaml
workflow:
  version: 1
  signals:
    id: NumberSignal
    results: UserSignal
    results2:
      - id: number
        name: string
  frames:
    - frame:
      view:
        component: OTUqiDataProvider
        props:
          source: trino-demo
          query: SELECT id, name FROM users
          outputs: [:id, :results, :results2]
```

I think I like this a lot. It declares the schema of the signal once instead of having an unnecessary abstraction in Record. This also feels the most terse yet expressive. The one slight downside is it moves the declaration of query results away from the query component, but I think this is fine because 1. this hopefully will never be hand written and 2. it should be clear that the outputs are :symbols and make it obvious to look at the signal store.

jonmagic: Yeah, I like it too 👍

hktouw: Great. I added the following as a todo to implement this.

> Update to use new signal annotation https://github.com/open-truss/open-truss/pull/128/files#r1518165179


I am going to clean up this PR, merge it then start working on this new annotation.

demo-app/src/open-truss/configs/8-mysql-demo.yaml
```diff
@@ -1,9 +1,21 @@
...
+  signals:
+    results: Records
   frames:
     - frame:
-      data:
-        query: select * from sys.metrics
-        source: mysql-demo
       view:
-        component: DataDemo
+        component: OTUqiDataProvider
```
hktouw, line 8: @jonmagic I updated the mysql demo to use the `OTUqiDataProvider` and everything just worked 🤘. I modified the kusto one too but didn't test it. I assume that should work as well.

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -247,7 +252,12 @@ function processProps({
       const signalName = parseSignalName(viewPropVal) ?? ''
       signal = signals[signalName]
     } else if (viewPropVal !== undefined) {
-      signal = signalsType.parse(viewPropVal)
+      // TODO This condition is for viewProps that are hard coded values (not signals)
+      // We need to transform them into a signal so that the component can use them.
+      // If creating many signals becomes an issue, perhaps we can make a simple wrapper
+      // object that mirrors the signal interface (eg. signal.value) and use that instead
+      signal = signalsType.parse(undefined)
```
hktouw, line 259: @kmcq the update to the signals code you made seem to break the code for passing in hardcoded view props. eg.

```yaml
props:
  title: Hello world
```

Previously we were doing `signal = signalsType.parse(viewPropVal)`, but the new validator logic you added causes this to throw an exception because hardcoded values fail `isSignalLike` as they are scalars / primitives. Instead parsing `undefined` gives us the default object and we just set the view prop value accordingly.

kmcq (approved):

PR #129, opened 2024-03-03
Use zod to specify signal shape

kmcq:
This is necessary for the config builder to know how to display signal types. This PR doesn't get it to work yet, but it gets us in the right direction.

After toiling for awhile to see if we could get zod to know the type of a signal, I went with changing the argument we had for default values to also specify the signal's value's shape and storing it alongside the signal in the global `SIGNALS` store. This feels correct to me because signals are props and we use zod to describe all props.

I also decided to export `z` which means OT applications are now pinned to OT's zod and don't need to install it themselves.


packages/open-truss/src/signals/index.ts
```diff
@@ -21,19 +36,25 @@ export function getSignalsType(
...
-  defaultValue: T,
-  validator?: () => boolean,
-): z.ZodDefault<z.ZodType<Signal<T>>> {
-  validator ||= () => true
+  valueShape: ValueShape<T>,
+): SignalsZodType<T> {
+  const defaultValue = valueShape.parse(undefined)
+  const validator = (val: unknown): boolean => {
+    return isSignalLike(val) && valueShape.parse(val.value) !== undefined
+  }
```
kmcq, line 50: Now that we have a zod shape for the signal's value, we can use it to validate that the correct values are passed in 🎉

packages/open-truss/src/signals/index.ts
```diff
@@ -21,19 +36,25 @@ export function getSignalsType(
...
   name: string,
-  defaultValue: T,
-  validator?: () => boolean,
-): z.ZodDefault<z.ZodType<Signal<T>>> {
-  validator ||= () => true
+  valueShape: ValueShape<T>,
+): SignalsZodType<T> {
+  const defaultValue = valueShape.parse(undefined)
+  const validator = (val: unknown): boolean => {
+    return isSignalLike(val) && valueShape.parse(val.value) !== undefined
```
hktouw, line 49: @kmcq Jon and I were reviewing this and were wondering if `valueShape.parse(val.value)` could ever be `undefined`?

hktouw (approved):
This looks great to @jonmagic and me! We left one small comment, but that shouldn't be blocking. We can merge this and make any follow up changes in post. Really excited to see zod being aware of signals and vice verse! I think that will be very powerful and give us more primitives to work with in the UI 💪

packages/open-truss/src/signals/index.ts
kmcq: Hmm I don't remember why I did this. I think we could just do this:

```suggestion
    return isSignalLike(val) && valueShape.parse(val.value)
```

PR #131, opened 2024-03-06
Add kusto-uqi-client and kusto-demo workflow

jonmagic:
## Why?

Closes https://github.com/open-truss/open-truss/issues/130

## How?

We wrote a uqi wrapper for the azure-kusto-data client and tested it by adding a kusto-demo workflow.

```yaml
workflow:
  version: 1
  frames:
    - frame:
      data:
        query: cluster("acme").database("widgets").RawEvent | limit 1
        source: kusto-demo
      view:
        component: DataDemo
```

NOTE: We tried to figure out how to run a kusto emulator locally with [these instructions](https://learn.microsoft.com/en-us/azure/data-explorer/kusto-emulator-install) but that ultimately failed because the linux docker image won't run on an M1 mac. So we had to configure the kusto demo settings in `.env.local` to point to a kusto cluster we use at work. Due to that I couldn't post a screenshot of the working demo.

## Collaborators

- @fionaochs 
- @tmetz


hktouw (approved):
Looks great to me! Nice work everyone! 🚀

Issue #132, opened 2024-03-08
Data Provider component follow ups


hktouw:
Follow ups from https://github.com/open-truss/open-truss/pull/128

- Update signals code to pass in signals to view props that are objects or arrays and not just strings. e.g. `output: [:results]`.
- Update to use new signal annotation https://github.com/open-truss/open-truss/pull/128/files#r1518165179
- Update component to only set the fields that the output has declared. Currently it just validates the fields exist in the results, but sets all the fields into the signal
- Support pagination
- Update component to set `isLoading` signal that can be used by display components to render loading indicators. Add `isLoading` to the base open truss component type so it's available in all components
- Update signals so that you can declare them at every frame level and signal name resolution moves up recursively up the frame parent tree. Similar to variable scoping in programming and how you can access variables in outer closures.
- Remove hardcoded UQI types in data component and instead import them from UQI

PR #133, opened 2024-03-10
Update renderer to dynamically generate signals

hktouw:
ref https://github.com/open-truss/open-truss/issues/132

### Why

We want a more ergonomic way to define signals and use them for data components. e.g. instead of designing signals in code we want to be able to define them in the config. This will make it easier to create ad-hoc schemas that don't need to be created in code. 

### How

This PR updates the renderer to dynamically generate signals. The renderer will create a zod signal object for Objects or Arrays with an object. e.g. signals defined in the yaml with the following shapes will be created.

```yaml
workflow:
  version: 1
  signals:
    resultsArray:
      - name: string
        phone: string
        acctbal: number
        comment: string
    resultsObject:
      name: string
      phone: string
      acctbal: number
      comment: string
```

This should be pretty straightforward to review commit by commit and by reading the commit comments.

@kmcq @jonmagic your reviews are requested


demo-app/src/open-truss/configs/4-trino-demo.yaml
```diff
@@ -12,16 +21,28 @@ workflow:
             SELECT name, phone, acctbal, comment
             FROM tpch.sf1.customer
             limit 10
-          results: :results
+          # output: [:resultsArray, :resultsObject, :name]
```
hktouw, line 24: I left this here to show what I am going to do as a follow up. I'm going to modify the renderer to pass in signals into objects and arrays as well as update the `OTUquiDataProvider` component to use the signals found in `outputs`. This will result in deleting `resultsArray` and `resultsObject` below.

jonmagic (approved):
Wow, was thinking about defining signals in workflows earlier today and was going to bring it up on Monday and viola you've already built it 😄 Great work!

I downloaded the branch and tested locally and looks great 👍

packages/open-truss/src/signals/index.ts
```diff
@@ -60,6 +63,60 @@ export function SignalType<T>(
...
+// - It adds the signal to the SIGNALS store using a sha256 hash of the object as the name
+// - It sets the default value of the signal to match the shape of the object or object[].
+//   string, boolean, and number  fields are set with default values of "", false, 0.
+// - It sets the type to unknown given that this is runtime and typescript types don't exist
+// - TODO it accpets predfined higher order types as valid types
+export function createSignal(signal: object | object[]): SignalsZodType {
+  const name = signalNameFromObject(signal)
+  if (name in SIGNALS) return SIGNALS[name].signal
+
+  const { zodShape, defValue } = createZodShape(signal)
```
kmcq, line 76: Can we spell out defaultValue?

packages/open-truss/src/signals/index.ts
```diff
@@ -60,6 +63,60 @@ export function SignalType<T>(
...
+  } else {
+    const zodSchema: Record<string, ZodTypeAny> = {}
+    const defValue: Record<
+      string,
+      string | boolean | number | object | object[]
+    > = {}
+
+    for (const [key, value] of Object.entries(signal)) {
+      if (isObject(value)) {
+        const { zodShape, defValue: defaultValue } = createZodShape(value)
```
kmcq, line 105: Thanks for handling recursion!

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -94,12 +100,15 @@ function createSignals(signalsConfig: SignalsV1, validate: boolean): Signals {
...
   Object.entries(signalsConfig).forEach(([name, val]) => {
-    const signal = SIGNALS[val]?.signal
-    if (validate && signal === undefined) {
-      const err = `Signal type ${String(
-        val,
-      )} is unknown. Please check value of ${String(name)} Signal`
-      throw new Error(err)
+    let signal: SignalsZodType | undefined
+    if (typeof val === 'string') {
+      signal = SIGNALS[val]?.signal
```
kmcq, line 105: Slight preference to move this after the if on the next line, make this if say "validate && !val in SIGNALS" and then the "let signal" above won't need the "| undefined".

kmcq (approved):
This turned out great! Thanks for doing this

hktouw:
@kmcq @jonmagic thanks for the feedback! Implemented your suggestions here! https://github.com/open-truss/open-truss/pull/133/commits/7f3b45335774e7ed9319e70971fcf5432258f302

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -94,14 +100,16 @@ function createSignals(signalsConfig: SignalsV1, validate: boolean): Signals {
   const signals: Signals = {}
   if (signalsConfig === undefined) return signals
   Object.entries(signalsConfig).forEach(([name, val]) => {
-    const signal = SIGNALS[val]?.signal
-    if (validate && signal === undefined) {
-      const err = `Signal type ${String(
-        val,
-      )} is unknown. Please check value of ${String(name)} Signal`
-      throw new Error(err)
+    let signal: SignalsZodType | undefined
```
hktouw, line 103: @kmcq unfortunately I had to keep the `undefined` here because this variable can be undefined if none of the conditions below matched.

PR #136, opened 2024-03-12
Improve renderer and OTUqiDataProvider for more ergonomic annotations

hktouw:
ref https://github.com/open-truss/open-truss/issues/132

### Why

We want to streamline the OTUqiDataProvider to have a more terse, ergonomic annotation and centralize schema logic to the signals. This will make writing and generating workflows much easier

### How

This change allows us to remove the following annotations:

```diff
workflow:
  version: 1
  signals:
    results:
      - Variable_name: string
  frames:
    - frame:
      view:
        component: OTUqiDataProvider
        props:
          source: mysql-demo
          query: select Variable_name from sys.metrics
+          output: [:results]
-          results: :results
-          output:
-            results:
-              - Variable_name: string
      frames:
        - frame:
          view:
            component: DataDemo
            props:
              results: :results

```

Additionally this adds support for having a scalar as a valid result type for `OTUqiDataProvider`. This will grab the first row of results from uqi and use the value that has the same key as the result type / signal name. See the third set of results below.

‎localhost:3000otplayground4-trino-demo 2024-03-11 at 8 37 11 PM

### Review

The easiest way to review this PR is commit by commit and reading the comments along the way.

@jonmagic @kmcq your reviews are requested
@fionaochs you mentioned you are going to be working more on Open Truss. You are more than welcome to review this PR as well if you are interested.


packages/open-truss/src/signals/index.ts
```diff
@@ -1,9 +1,13 @@
 import { z, type ZodTypeAny } from 'zod'
-import { signal, type Signal } from '@preact/signals-react'
+import { signal, type Signal as PreactSignal } from '@preact/signals-react'
 import { typeToZodMap, typeToDefaultValue } from '../utils/describe-zod'
 import { isObject } from '../utils/misc'
 import CryptoJS from 'crypto-js'
-export { Signal }
+
+export interface Signal<T = any> extends PreactSignal<T> {
```
kmcq, line 7: This PR is a bit too substantive for me to review on my phone but a quick question: could/should this yaml name thing be added alongside signal and valueShape (define a few lines below) instead of directly on the signal?

kmcq (commented):
Just a quick comment. I'll read more in about a week

fionaochs:
Tested this out locally and got this error which traced to https://github.com/open-truss/open-truss/blob/91d657bd030d502352200d605e9db962b243aca5/packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx#L162-L163 which @jonmagic could not reproduce. 
 
image

Testing in console, you cannot `JSON.parse` an empty string, but parsing `null` would return `null`
image

packages/open-truss/src/signals/index.ts
```diff
@@ -28,6 +32,10 @@ export function getSignalsType(
   return getSignalAndValueShape(possibleZodObject)?.signal
 }
 
+export function getSignalsShapeFromSignal(signal: Signal): SignalsZodType {
```
kmcq, line 35: How about `signalValueShape`? It's shorter and mentions `valueShape` which I think is a good thing

packages/open-truss/src/uqi/uqi.ts
```diff
@@ -8,6 +8,78 @@ export type UqiMappedType =
...
+  return result.rows.map((row) => {
+    const rowResult: Record<string, unknown> = {}
+    row.values.forEach(({ key, type, value }) => {
+      rowResult[key] = castUqiValue(type, value)
+    })
+    return rowResult
+  })
+}
+
+export function castUqiValue(type: string, value: string): unknown {
```
kmcq, line 40: Could we move this up next to `UqiMappedType`'s definition (line 3)? And could `type` be typed as `UqiMappedType` here?

```suggestion
export function castUqiValue(type: UqiMappedType, value: string): unknown {
```

packages/open-truss/src/signals/index.ts
kmcq: Ah okay I get what's going on here: inside the component, we don't have access to the full object stored in `SIGNALS`, so we have to add this property to the signal itself. I'll think on this a bit

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -278,6 +271,48 @@ function processProps({
...
+      newProps[propName] = processViewProps(subProps)
+    } else if (Array.isArray(prop)) {
+      const propsFromArray = prop.map((value) => {
+        const tmpKey = '_key'
+        const subViewProps = { viewProps: { [tmpKey]: value } }
+        const subProps = Object.assign({}, props, subViewProps)
+        return processViewProps(subProps)[tmpKey]
+      })
+
+      newProps[propName] = propsFromArray
```
kmcq, line 308: > Updates the processProps code to recursively parse through the view props and set any signals that exist in nested arrays or objects

I love me some recursion but I'm not sure this is worth it 🤔 What led you to add this? It's not very complicated so maybe it's fine to keep, but I find us handling YAML objects arrays, which it does natively, by iterating over things to be a smell.

packages/open-truss/src/components/OTUqiDataProvider.tsx
```diff
@@ -63,72 +42,22 @@ const UqiDataProvider: BaseOpenTrussComponentV1<z.infer<typeof Props>> = (
...
+    const parsedResults = parseUqiResult(queryResults)
+    const defaultValue = shape.parse(undefined)
 
-  return children
-}
+    const result = Array.isArray(defaultValue)
+      ? parsedResults
+      : isObject(defaultValue)
+      ? parsedResults[0]
+      : parsedResults[0]?.[signal.yamlName]
```
kmcq, line 54: This is clever! I missed this in the last PR, but could you say more about returning single results?

I think I would suggest that we only do array outputs and thus only accept array signals for `output`. It would keep this simpler and more closely aligns this to UQI, which I think always returns multiple results. It also means we don't need the `signal.yamlName` field.

This would also set us up better to in the future let folks do more custom map-reduce functions: the data provider would always pass along an array that can be acted upon.

If it's really helpful to pluck just the first element, I'd suggest we do that by passing a prop to this data provider instead of only through the definition of the signal we're sending the result to.

If we do what I'm saying, your examples could then look like the following, noting that the `resultsObject` is removed as it's equivalent to `resultsArray`:

```yaml
  signals:
    resultsArray: # OTUqiDataProvider returns collection with rows using the shape of the object
      - name: string
        phone: string
        acctbal: number
        comment: string
    # this maps rows to just `[{ name: "row1Name" }, { name: "row2Name" }, ...]`
    name:
      - name: string
```

packages/open-truss/src/components/OTUqiDataProvider.tsx
```diff
@@ -63,72 +42,22 @@ const UqiDataProvider: BaseOpenTrussComponentV1<z.infer<typeof Props>> = (
...
-  metadata: UqiColumn[],
-): boolean {
-  // Define a mapping from the types in metadata to JavaScript's typeof values
-  const typeMap: Record<string, string> = {
-    String: 'string',
-    Number: 'number',
-    Boolean: 'boolean',
-    BigInt: 'bigint',
-    // Add more mappings as necessary
+    const validatedResult = shape.parse(result)
```
kmcq, line 56: Yes! This is much better 🎉

kmcq (commented):
Looking really cool, I have an idea that we should discuss!

packages/open-truss/src/components/OTUqiDataProvider.tsx
jonmagic: We made this change to support signal shapes that only have one value instead of an array of values. That way you can re-use display components without having to change the component signature or have to introduce another nested frames/frame that returns the first element.

We can change this if there continue to be strong feels ❤️

packages/open-truss/src/configuration/engine-v1/Frame.tsx
jonmagic: > but I find us handling YAML objects arrays, which it does natively

What did you mean by "it does natively" here?

packages/open-truss/src/signals/index.ts
jonmagic: 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔

packages/open-truss/src/signals/index.ts
jonmagic: 9031dca680437f7f701cebf779faffa2ce82b413

packages/open-truss/src/uqi/uqi.ts
jonmagic: 2e166dca9b366be9f57ab90d732b6b4662b99853

jonmagic (approved):
I'm approving with the caveat that we may return to some comments that @kmcq left once we have a chance to speak with him again ❤️

PR #146, opened 2024-03-26
Add build step and dev tools for adding `.js` and `.jsx` extensions to file imports

hktouw:
## Why

Javascript runtime targets often need file imports to have the correct extensions. e.g. `import file from './file.js'`. Currently OT is outputting mjs files without extensions. e.g. `import file from './file'`. We should update our project so that we are outputting correctly formatted import statements

## How

This PR adds a new script that processes our `dist/mjs` files and adds `.js` and `.jsx` extensions to every file import. Additionally it updates our build steps and local dev tools to run this script when we run `script/server` and when we publish the package.

Paired with @jonmagic on this.


jonmagic (approved):
Works as advertised 🤘

PR #147, opened 2024-04-03
Update parseYaml function to take optional generic type

jonmagic:
## Why?

Because we ran into a situation where we wanted this and it did not support it.

**PLEASE NOTE** this does not actually do runtime type checking on the parsed version of the yaml string being parsed. If we want that we should incorporate zod into this function.

Co-authored-by @spicycode, @tmetz, @hktouw, @andrew-dillon, @fionaochs


hktouw (approved):
🥳

kmcq (approved):
Legit

PR #152, opened 2024-04-11
Fix some bugs and add data demo

jonmagic:


packages/open-truss/src/components/OTUqiDataProvider.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react'
```
hktouw, line 1: Worth noting that in @kmcq's setup he doesn't have the preset react loader in his webpack config, so he needs this import to be there. For now I think we can drop this, but perhaps @kmcq can look into getting the preset react loader added to his webpack setup when he moves to babel. Alternatively, we can add another post build script to manually add the react imports, similar to how we are adding `.js` extensions to files.

hktouw (approved):
👍

PR #162, opened 2024-04-18
Force query execution by incrementing force_query NumberSignal

jonmagic:
## Why?

We want to be able to re-run a query if we know the table it is querying has been updated.

## How?

I had an idea to use a counter NumberSignal that you can increment in your application code and the OTUqiDataProvider can trigger execution of the query when the counter changes. 

Here's an example of a workflow that uses it:

```yaml
workflow:
  version: 1
  signals:
    query: string
    force_query: number
    results:
      - id: number
        value: string
        description: string
        created_at: string
        creator_id: number
        datasource_name: string
  frames:
    - frame:
      view:
        component: Input
        props:
          query: :query
          template: >
            SELECT
              datasource_entries.id,
              datasource_entries.\`value\`,
              datasource_entries.description,
              DATE_FORMAT(datasource_entries.created_at, '%Y-%m-%dT%T+00:00') AS created_at,
              datasource_entries.creator_id,
              datasources.name AS datasource_name
            FROM datasource_entries
            JOIN datasources ON datasource_entries.datasource_id = datasources.id
            WHERE datasource_entries.datasource_id = 17
            AND (datasource_entries.value LIKE :value OR datasource_entries.description LIKE :value)
          default_query: >
            SELECT
              datasource_entries.id,
              datasource_entries.\`value\`,
              datasource_entries.description,
              DATE_FORMAT(datasource_entries.created_at, '%Y-%m-%dT%T+00:00') AS created_at,
              datasource_entries.creator_id,
              datasources.name AS datasource_name
            FROM datasource_entries
            JOIN datasources ON datasource_entries.datasource_id = datasources.id
            WHERE datasource_entries.datasource_id = 17
    - frame:
      view:
        component: OTUqiDataProvider
        props:
          source: spamurai
          query: :query
          force_query: :force_query
          output: [:results]
      frames:
        - frame:
          view:
            component: Datasource
            props:
              results: :results
              force_query: :force_query
```

And then the Datasource component takes in the signal:

```ts
  ...withChildren,
  results: UnknownSignal,
  force_query: NumberSignal,
})
```

And when the component needs the query to execute it again it can just do this:

```ts
force_query.value += 1
```


jonmagic:
Been thinking through other ways to accomplish this. I believe @hktouw suggested exposing a function as a signal that you could call from your application to force the component to rerun the query.

What other ways could we implement it? What other APIs could we leverage? What feels the most ergonomic? @open-truss/engineers

kmcq (approved):
I like it. I wonder if we should build this into the rendering engine so you don't have to do the extraneous header. Maybe something that increments a component's `key`. That's just a thought tho this is great

andrew-dillon:
I like exposing this feature via a `rerender` function.

hktouw (approved):

hktouw:
> I wonder if we should build this into the rendering engine so you don't have to do the extraneous header. 

That's a really interesting idea. I think we probably do want the general ability to re-render any component and make that accessible by any component. I don't know what an ergonomic config for that might be. 

I'm going to merge this for now since it's working, but let's keep an eye out for more use cases like this and think through how best we want to expose these "meta" functions to interact with the workflow engine.

PR #163, opened 2024-04-19
Expose more of @preact/react-signals and make switching apps easier

jonmagic:
## Why?

I would like to have access to all of the public API from the `@preact/signals-react` library. I also need to switch between development on my work app and the demo-app easier. We already had things working on the work app but when switching back to the demo-app it wasn't fixing the npm link for react.

## How?

See commits.


andrew-dillon (approved):

hktouw (approved):
👍

PR #165, opened 2024-04-24
Add debug flag

hktouw:
To make debugging easier in OT components, this PR extends the base OT component type with a `debug` prop. OT components can then opt into using that prop to output useful debugging results. Debugging is then enabled by setting `debug: true` at the top of the config file. Later, this may wanted to be set elsewhere, but this felt like the most sense for the time being.

You can see in this demo, debugging is added to the `OTUquiDataprovider` component.

https://github.com/open-truss/open-truss/assets/4481584/f52cf631-8b07-473d-ab6a-ca9c68a77eb0

/cc @open-truss/engineers

### DIFF
--- a/demo-app/src/open-truss/configs/4-trino-demo.yaml
+++ b/demo-app/src/open-truss/configs/4-trino-demo.yaml
@@ -1,5 +1,6 @@
 workflow:
   version: 1
+  debug: true
   signals:
     resultsArray: # OTUqiDataProvider returns collection with rows using the shape of the object
       - name: string
--- a/packages/open-truss/src/components/OTUqiDataProvider.tsx
+++ b/packages/open-truss/src/components/OTUqiDataProvider.tsx
@@ -23,9 +23,10 @@ export const Props = BaseOpenTrussComponentV1PropsShape.extend({
 const OTUqiDataProvider: BaseOpenTrussComponentV1<z.infer<typeof Props>> = (
   props,
 ) => {
-  const { query, children, output, source } = props
+  const { query, children, output, source, debug } = props
 
   useSignalEffect(() => {
+    if (debug) console.log({ query: query.value, source: source.value })
     let queryResults: SynchronousQueryResult
     if (query.value === '') return
     const fetchData = async function (): Promise<undefined> {
@@ -35,6 +36,7 @@ const OTUqiDataProvider: BaseOpenTrussComponentV1<z.infer<typeof Props>> = (
         body: JSON.stringify({ query, source }),
       })
       const deserialized = await result.json()
+      if (debug) console.log({ api_response: result })
       queryResults = deserialized
     }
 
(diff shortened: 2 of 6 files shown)


kmcq (commented):
I like it! I think it should have a more unique name though. Maybe `__OT_DEBUG__`?

demo-app/src/open-truss/configs/4-trino-demo.yaml
```diff
@@ -1,5 +1,6 @@
 workflow:
   version: 1
+  debug: true
```
spicycode, line 3: I was thinking when reading it I'd normally see my LSP docs say "Debug is used to automatically output debug info to the console when enabled" or something like that.  

Do we have a schema defined yet for this? I'm wondering what we're thinking around the idea of documenting this config options.

packages/open-truss/src/components/OTUqiDataProvider.tsx
```diff
@@ -54,6 +56,12 @@ const OTUqiDataProvider: BaseOpenTrussComponentV1<z.infer<typeof Props>> = (
             : parsedResults[0]?.[signal.yamlName]
 
           const validatedResult = shape.parse(result)
+          if (debug)
```
spicycode, line 59: If it ever makes sense (aka there are no early returns in a given function with debug lines) we could do something like 

```typescript
// near start of function
if (debug) console.group('MeaningfulNameHere')
...

// near end of function
if (debug) console.groupEnd()
```

Which would make the debug data be grouped by whatever we choose, and ease reading it a bit.

spicycode (approved):

packages/open-truss/src/components/OTUqiDataProvider.tsx
hktouw: TIL, really cool.

When playing around with it using the [w3schools example](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_console_group), I noticed calling group without groupEnd created multiple nesting levels. I'm guessing that's probably what you are referring to as when we'd only want to use this when there are not returns.

W3Schools Tryit Editor 2024-04-25 at 9 53 53 AM

I _think_ my preference might not be to use this pattern as if there is any failure in the code before groupEnd is called, it could cause infinite levels of nesting and make the output messy. But if that's what folks want to use I don't feel strongly against using it in general.

demo-app/src/open-truss/configs/4-trino-demo.yaml
hktouw: We don't have the schema documented anywhere, but agreed. We'll probably get to the point soon where we'll want docs for this.

packages/open-truss/src/components/OTUqiDataProvider.tsx
spicycode: > I think my preference might not be to use this pattern as if there is any failure in the code before groupEnd is called

I think that's very reasonable.  Often the way I see libraries get around this is by effectively having a defer in a promise chain (so always run this at the end, regardless of error or success), and I feel like that is way too much for the value we'd get out of it now.

hktouw:
@kmcq how does https://github.com/open-truss/open-truss/pull/165/commits/7a4fb591159ed2d0977719a62adcb86bdcb86cff look to you? I went with `_DEBUG_` to the variable name is shorter, but it should still be pretty unique as a prop name.

kmcq (approved):
Looks good to meeee

PR #166, opened 2024-04-27
build: Don't preserve JSX for ESM builds

kmcq:
## Why?

When I installed Open Truss in an application that didn't have a JSX transform as part of its webpack configuration it broke in two ways:

1. It didn't know how to handle `.jsx` files.
2. It failed to build files that had JSX syntax in them but didn't have `import React from 'react'`.

I believe not preserving JSX for our ESM build will solve both of these issues:

1. We won't be keeping `.jsx` files.
2. With the newer `react-jsx` parser, [React is no longer needed to be in scope to parse JSX](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).

## How?

I just deleted that option from the ESM configuration. Note that we have `"jsx": "react-jsx"` in our main `tsconfig.json` file so the ESM version will inherit that value.

### Testing

- Demo app works.
- GitHub folks it would be great if you could try this in your app. (cc @hktouw, knower of all things npm/ts build systems)


kmcq:
> 2\. With the newer `react-jsx` parser, [React is no longer needed to be in scope to parse JSX](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).

FWIW I was really happy to see this as I didn't want to add an eslint rule that would force us to always add the import :)

hktouw:
Hmm, it doesn't seem to work on the demo app. I need to head out, but I'll come back to this later.

hktouw:
Running into an issue where there is a conflict with the following code outputted by the ts compiler. This is outed by the `"jsx": "react-jsx"` we set in our `tsconfig`.

```js
 import { jsx as _jsx, jsxs as _jsxs } from 'react/jsx-runtime.js'
```

When using the same settings for the demo-app and our app

```js
    "module": "esnext",
    "moduleResolution": "bundler",
```

The demo app complains if `'react/jsx-runtime.js'` has a `.js` extension, whereas our app complains when it doesn't have a `.js` extension. 

@kmcq when you try installing this in your app, does your setup want or not want the `.js` extension?

hktouw:
I found this [webpack config](https://webpack.js.org/configuration/resolve/#resolveextensions) that tells webpack to ignore import statements that don't have file extensions. This didn't work, but then I noticed this line added to the error output

> (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"')

In our build setup we are [outputting](https://github.com/open-truss/open-truss/blob/main/packages/open-truss/bin/append-package-json-to-dist) a package.json file with `'type': 'module'` in our `dist/mjs` directory. When I removed that however, this started working in both apps. I recall this `package.json` file being necessary at the time, but don't remember why. @jonmagic do you remember? Perhaps this was necessary at the time, but we've updated our configs so much maybe our current setup makes it no longer necessary? 

I think the demo app and our app setups don't actually require extensions anymore if we remove the `package.json` file which was forcing the requirement. This is because our apps are setup to know how to resolve extension less imports. @kmcq can you see if your app setup has that webpack config?

hktouw:
Javascript is the worst

kmcq:
@hktouw thanks for investigating this! And I'm glad to see you were able to get things to work without the extra scripts 🎉 

I tried to figure out how to get my app to install by using a `package.json` `file:` entry but didn't have any luck -- its setup is complicated and doesn't seem to support it (easily at least). If this works for you all, could we ship it and I try by installing from npm?

hktouw (approved):
Let's ship this so you can give this a try 👍. Worst case we'll cut a new PR if there are any bugs.

PR #167, opened 2024-04-30
Css pipeline POC

hktouw:
### Why

Currently we [can't share front-end display code](https://github.com/open-truss/open-truss/discussions/161#discussioncomment-9216950) (e.g. config builder) [share front-end code](https://github.com/open-truss/open-truss/discussions/161#discussioncomment-9216950). We should address this issue so that we can more easily collaborate on the config builder and other OT pages.

### How

This PR is a POC for a workflow where we build entire React pages in the OT package, generate css files, and import the JS and CSS into applications. If you pull down this branch you can see this demo'd on http://localhost:3000/ot/playground/11-css-import. There is an OT page that is merely an OT component that adds a green background, we import that OT component and css into the demo-app, and we render that with a child so you an see the green background.

‎localhost:3000otplayground11-css-import 2024-04-29 at 5 49 12 PM

### Considerations

- I had to install another build tool (rollup) in order to compile the css (we could use scss too). I updated our dev scripts to run this automatically as you would expect. e.g. `npm run build` will generate the js and css assets and `npm run tsc:watch` will run the the js and css compilers automatically when there are changes to any files. When briefly testing, the dev experience seemed fine and as you would expect. I have yet to test this in our app.
- Unfortunately, I had to install rollup as a dev package tool to the demo-app package.json. This adds the rollup binary which is used to run the build scripts. Not ideal to have that as a dev dependency, but hopefully that won't be an issue. 
- I would appreciate people's feedbacks and thoughts on this approach when you have a moment.

cc @jonmagic @kmcq since we were talking about this in the discussion


kmcq:
@hktouw Can you say more about why you had to install rollup? I haven't done any research into this yet but was hoping we'd be able to just specify in package.json's "files" section that we want to include css files in our package.

hktouw:
Mostly so that we can combine all the css files into one, so applications don't have to import multiple files and can just add it to the root. Alternatively, if we want to go the route of having each component on the app side import their own css files I think that would work just as fine. That's probably a better route in the short term. I can update this PR to drop rollup and do that instead. Later if we have too many css files or want to use a system like sass we can re-add this.

jonmagic (approved):

jonmagic:
Let's see if we can get this working in the day job apps now. Merging.

Issue #169, opened 2024-05-02
Implement StoredQuery API in the demo-app


jonmagic:
Use sqlite for the database and graphql for the API and it should have these resolvers:

- createStoredQuery
  - args
    - query
    - source
    - name
    - description
- updateStoredQuery
  - args
    - id
    - name
    - description
    - expireAt
- archiveStoredQuery
  - args
    - id
- showStoredQuery
  - args
    - id
  - fields
    - id
    - query
    - source
      - name
      - type (mysql, trino, kusto, etc)
    - creatorId
    - createdAt
    - startedAt
    - completedAt
    - archivedAt
    - expireAt
    - failedAt
    - failedReason
    - columns
    - tables
- showStoredQueryRows
  - args
    - first
    - after
    - before
    - last
  - fields
    - edges
      - node
        - seializedData
- searchStoredQueries
  - args
    - first
    - after
    - before
    - last
    - filterBy
      - creatorId
      - serializedColumns
      - serializedTableNames
      - started
      - completed
      - failed
      - archived
  - fields
    - edges
      - node (StoredQuery)

jonmagic:
I think this can be closed.

PR #174, opened 2024-05-13
Example sqlite-uqi-client for the demo-app

jonmagic:
## Why?

I'm working on a branch for https://github.com/open-truss/open-truss/issues/169 and wanted an sqlite-uqi-client to use.

## How?

1. Added `sqlite3` and `@types/sqlite3` dependencies.
2. Created an empty sqlite db: `cd demo-app && mkdir -p src/db && sqlite3 src/dev.db`.
3. Implemented `sqlite-uqi-client`.
4. Drove the implementation by developing a simple workflow.
5. Fixed a number of typescript types issues.

### DIFF
--- a/demo-app/src/lib/sqlite-uqi-client.ts
+++ b/demo-app/src/lib/sqlite-uqi-client.ts
@@ -1,0 +1,112 @@
+import {
+  Iterator,
+  uqi,
+  type UqiScalar,
+  type UqiClient,
+  type UqiColumn,
+  type UqiContext,
+  type UqiMappedType,
+  type UqiResult,
+} from '@open-truss/open-truss'
+import sqlite from 'sqlite3'
+import { URL } from 'url'
+
+export interface SqliteConfig {
+  uri: string
+}
+
+interface AsyncSqliteClient {
+  query: (
+    query: string,
+  ) => Promise<[Array<Record<string, UqiScalar>>, UqiColumn[]]>
+  teardown: () => Promise<void>
+}
+
+async function setupAsyncSqliteClient(uri: string): Promise<AsyncSqliteClient> {
+  const parsedUrl = new URL(uri)
+
+  return new Promise((resolve, reject) => {
+    const db = new sqlite.Database(parsedUrl.pathname, (error) => {
+      const og = {
+        query: async (query: string): Promise<[any[], UqiColumn[]]> => {
+          return new Promise((resolve, reject) => {
+            db.all<Record<string, UqiScalar>>(query, (error, rows) => {
+              if (error) {
+                reject(error)
+              } else {
+                const columns = rows.length > 0 ? columnsForRow(rows?.[0]) : []
+                resolve([rows, columns])
+              }
+            })
+          })
+        },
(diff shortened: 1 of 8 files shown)


demo-app/src/lib/sqlite-uqi-client.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+  })
+}
+
+export default async function (config: SqliteConfig): Promise<UqiClient> {
+  async function query(
+    context: UqiContext<SqliteConfig, AsyncSqliteClient>,
+    query: string,
+  ): Promise<AsyncIterableIterator<UqiResult>> {
+    const [rows, columns] = await context.client.query(query)
```
jonmagic, line 85: This currently returns all rows immediately. This could be updated to more of a streaming model like we use for kusto/mysql/trino but I wanted to ship something quickly for now and typically an sqlite database is going to be smaller and used for development.

demo-app/src/lib/uqi-sources.ts
```diff
@@ -1,8 +1,23 @@
 import createKustoUqiClient from '@/lib/kusto-uqi-client'
 import createMysqlUqiClient from '@/lib/mysql-uqi-client'
+import createSqliteUqiClient from '@/lib/sqlite-uqi-client'
 import createTrinoUqiClient from '@/lib/trino-uqi-client'
+import { type UqiClient } from '@open-truss/open-truss'
 
-const sources = {
+type SourceConfig = any // TODO: make a type that can represent any source config
```
jonmagic, line 7: Ran out of time getting the types for this right in my time-boxed session so I added. TODO for later.

demo-app/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "ES2018",
```
jonmagic, line 3: Had to change the target to fix a warning about capture groups not working on earlier versions here: https://github.com/open-truss/open-truss/blob/2277e1a39e73079836bf9d97ce74a791bff6e493/demo-app/src/lib/kusto-uqi-client.ts#L36

demo-app/src/lib/sqlite-uqi-client.ts
spicycode: I like the idea of getting something out more quickly too.  I've definitely seen giant sqlite db's (mainly on macOS via CoreData), but in web apps I think this would be so fast even if it was a large db that no one will notice.

jonmagic:
Closing this for now after discovering that sqlite doesn't use strict types, everything is dynamic types.

You can use introspection queries to discover the types of fields based on the CREATE TABLE statement but you can't definitively determine the type of a column in a SELECT because you might modify the column output with a function, for example parsing a string as a number or counting the length of a string to get it's length. These computed columns don't have a definitive type that you can rely on.

Another hack would be to wait for all rows to return and then iterate through all of them to build a map of the most "valid" types for each column, but this will be prone to bugs in a system like open-truss where we rely on accurate types.

PR #175, opened 2024-05-14
Update formatQuery to support more types

jonmagic:
## Why?

Because I ran into an issue while working on https://github.com/open-truss/open-truss/issues/169


hktouw (approved):
Looks great to me 👍

PR #171, opened 2024-05-09
Add graphql api to demo-app

jonmagic:
## Why?

I want to start working on https://github.com/open-truss/open-truss/issues/169 but first wanted to add a simple graphql API to the demo-app that supports query, mutation, and subscription and also uses codegen to generate types from the graphql SDL and examples.

## How?

I leveraged ChatGPT heavily but also referenced some docs when ChatGPT gave me old examples

- https://chat.openai.com/share/0228ac85-db19-4cf8-baff-3167c1315517
- https://github.com/dotansimha/graphql-code-generator/issues/1090
- https://the-guild.dev/graphql/yoga-server/docs/features/subscriptions
- https://chatgpt.com/c/c76cff32-af69-4875-8d4a-83bc201892b6

Run the server:

```bash
script/server
```

Open the graphiql UI: http://localhost:3000/api/graphql

Examples:

```graphql
query {
  hello {
    message
  }
}
```

```graphql
mutation {
  hello(input: {name: "jon"}) {
    message
  }
}
```

```graphql
subscription {
  hello(names: ["andrew-dillon", "demiracl", "fionaochs", "hktouw", "jonmagic", "kmcq", "kre64", "ndanny", "tmetz"]) {
    message
  }
}
```

PR #176, opened 2024-05-14
Reorganize graphql schema

jonmagic:
## Why?

I'm working on https://github.com/open-truss/open-truss/issues/169 and wanted to reorganize the graphql schema so that Mutation, Query, and Subscription each have their own folders and each function used in them each gets their own file.

You can test this out by:

```bash
cd demo-app
script/server
```

Then open http://localhost:3000/api/graphql and try the same examples from https://github.com/open-truss/open-truss/pull/171


demo-app/src/graphql/Subscription/hello.ts
```diff
@@ -0,0 +1,17 @@
+import sleep from '@/lib/sleep'
+import {
+  type HelloResponse,
+  type SubscriptionHelloArgs,
+} from '@/graphql/generated-types'
+
+const hello = {
+  subscribe: async function* (_: unknown, { names }: SubscriptionHelloArgs) {
```
hktouw, line 8: Oh interesting. You want to support subscriptions? At of curiosity, do you have any plans to make a subscription?

hktouw (approved):
All this scaffolding seems reasonable to me 👍

demo-app/src/graphql/Subscription/hello.ts
spicycode: We use subscriptions at work already for a few things, so I imagine it might be an adaption of our existing usages but within open truss.

spicycode (approved):

demo-app/src/graphql/Subscription/hello.ts
jonmagic: Yeah, I imagine there will be use cases at some point so I figured I would support it from day one if it wasn't too much trouble. That's one of the reasons I chose graphql-yoga as it comes with support and was able to get it to work with nextjs.

PR #180, opened 2024-05-15
Update to latest nodejs 20 version

jonmagic:
## Why?

I noticed we were still on nodejs 18.

## How?

Updated `.tool-versions` files.


spicycode:
I think you'll also want to update ci-checks.yml to set a 'node-version-file' (see https://github.com/actions/setup-node?tab=readme-ov-file#usage) so CI will stay in sync going forward.

spicycode (approved):

hktouw (approved):

PR #182, opened 2024-05-16
Make signals nullable

hktouw:
## Why

Per our discussion in Discord, we want to make signal values (not signals) nullable because there are cases where apis and datasources will return null instead of their stated values. e.g. @jonmagic encountered a scenario where a column in mysql is nullable and was returning null.

We should update signals to be nullable and make components handle the null case.

## How

This PR
- updates the signals code to make all signal values nullable and consider null as valid instead of throwing an exception. This will also help us disambuguate between not provided values vs their zero values (e.g. `0`, `''`). We were previously casting values to their zero value and had no way of knowing if no value was supplied or if the value was the zero value.
  - ‎localhost3000otplayground14-nullable-fields 2024-05-16 at 5 13 45 PM
- updates the uqi backend code to not cast values to strings when sending them over the wire. This gives us access to types like `null` instead of having them casted to `"null"`.
- Adds more useful debugging logs
  - ‎localhost3000otplayground14-nullable-fields 2024-05-16 at 5 12 58 PM
- Throws an exception when a signal is used by a component and not declared as a signal. This was causing an inscrutable error before.
  - ‎localhost3000otplayground14-nullable-fields 2024-05-16 at 5 01 05 PM
- Adds more useful information to zod parsing errors for signals.
- Updates the describeZod code to have an `isNullable` field.

The easiest way to review this PR is commit by commit

### Nullable nested Array and Object values

This PR updates the code to make nested values nullable for dyanamic signals (e.g. signals defined only in yaml). I did not do this for predefined signals since it wasn't straightforward to do. How we would have to do this is either:
- use the describeZod code to get the zod object tree and recreate the zod object with each field being nullable.
- change predefined signals to not pass in zod objects, but normal objects instead of which we can use to create the zod object.

For now this will only be an issue if predefined signals are being populated by dataproviders with potentiall null values. WE'll likely come across this sooner or later and when this occurs we can do one of the solutions above.

/cc @open-truss/engineers


jonmagic (approved):
This looks great and works great! Thank you so much @hktouw 🙏

PR #183, opened 2024-05-16
Improve support for Array signals

jonmagic:
## Why?

We need signals to support array values because those are common return types and data structures in databases

```yaml
workflow:
  version: 1
  signals:
    r1:
      value: number[]
    r2:
      value: string[]
  frames:
    - frame:
      view:
        component: OTUqiDataProvider
        props:
          source: mysql-demo
          query: >
            SELECT JSON_ARRAY(1, NULL, 3) AS value
          output: [:r1]
      frames:
        - frame:
          view:
            component: DataDemo
            props:
              results: :r1
    - frame:
      view:
        component: OTUqiDataProvider
        props:
          source: mysql-demo
          query: >
            SELECT JSON_ARRAY('one', NULL, 'three') AS value
          output: [:r2]
      frames:
        - frame:
          view:
            component: DataDemo
            props:
              results: :r2
```

## How

This PR
- https://github.com/open-truss/open-truss/pull/183/commits/7cdf5873c3942c80a59ff1e06f48df411cc84f50 Updates the dynamic signals code to do lookups in the pre-existing list of signals. Previously, when you created a signal that wasn't in the predefined list, it would not use that list when defining signals inside of nested types (e.g. object properties). Now you can use predefined signals in nested types.
-  https://github.com/open-truss/open-truss/commit/94919ce621bd767ed8e3250918f1aec8fced2b00 Updated the front-end uqi type casting code to not call JSON.parse when the return value is a JSON type and it is already an array or object. The value has already been parsed and it was throwing an exception trying to parse it again.
- https://github.com/open-truss/open-truss/commit/e1d20bc3f395be001d978a6fb51167e3d8926854 Adds an array type mapping for the trino client so that uqi can process array types from trino.
- https://github.com/open-truss/open-truss/commit/599683875b87488e4560adaf84ea25ba0b70528c Cleans up the dynamic signals code to be more legible and adds support for multi-dimensional arrays.
- https://github.com/open-truss/open-truss/commit/5cb1d2e7271e3e5594347565ce0170560c62e45a Add demo for alternate multi dimensional array syntax 

cc @open-truss/engineers

### DIFF
--- a/packages/open-truss/src/signals/index.tsx
+++ b/packages/open-truss/src/signals/index.tsx
@@ -171,17 +171,17 @@ export const NavigateFrameSignal = SignalType<NavigateFrame>(
 export type NavigateFrameSignalType = z.infer<typeof NavigateFrameSignal>
 
 // Scalar types
-export const NumbersSignal = SignalType<(number | null)[]>(
+export const NumbersSignal = SignalType<Array<number | null>>(
   'number[]',
   z.array(z.number().nullable()).default([]),
 )
 export const NumberSignal = SignalType<number>('number', z.number().default(0))
-export const StringsSignal = SignalType<(string | null)[]>(
+export const StringsSignal = SignalType<Array<string | null>>(
   'string[]',
   z.array(z.string().nullable()).default([]),
 )
 export const StringSignal = SignalType<string>('string', z.string().default(''))
-export const BooleansSignal = SignalType<(boolean | null)[]>(
+export const BooleansSignal = SignalType<Array<boolean | null>>(
   'boolean[]',
   z.array(z.boolean().nullable()).default([]),
 )
--- a/packages/open-truss/src/signals/index.tsx
+++ b/packages/open-truss/src/signals/index.tsx
@@ -106,7 +106,7 @@ export function SignalType<T>(
 // - TODO it accepts predefined higher order types as valid types
 export function createSignal(
   configName: string,
-  signal: object | object[],
+  signal: object,
 ): SignalsZodType {
   const name = `${configName}-${signalNameFromObject(signal)}`
   if (name in SIGNALS) return SIGNALS[name].signal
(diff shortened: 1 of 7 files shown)


hktouw:
> SELECT JSON_ARRAY(1, NULL, 3) AS value

I think this returns a string since it is json so that is a type mismatch with the signal. The signal should be of type `string` instead of `number[]` or `string[]`.

That being said, dynamic signals does not yet support arrays or using predefined signals. So it doesn't support:

```
  signals:
    r1:
      arrayValue: number[]
    r2:
      predefinedSignal: User
```

Since we do have `number[]`, `string[]`, `boolean[]` as a predefined signals, all we need to do is support using predefined signals inside of dynamic signals. This commit does that and also updates the query in your workflow to return an array type instead of json string. https://github.com/open-truss/open-truss/pull/183/commits/7cdf5873c3942c80a59ff1e06f48df411cc84f50

However, it does not seem that uqi knows how to return array types as I'm getting this following error. I think once we teach uqi to know how to take array results from the database and convert them into json arrays we should be good to go. 

‎localhost3000otplayground15-arrays-signals-from-query 2024-05-17 at 6 20 37 PM

As an aside, I added the [debug flag](https://github.com/open-truss/open-truss/pull/183/commits/5c7a0ac7040c740ef6a3a1477eb63a5ea718da8a) to the workflow. Highly recommend keeping this on as the OT data provider component has a lot of useful output with that flag on.

jonmagic:
Thank you @hktouw!

> However, it does not seem that uqi knows how to return array types as I'm getting this following error. I think once we teach uqi to know how to take array results from the database and convert them into json arrays we should be good to go.

Since mysql does support a JSON type and uqi supports a JSON type we just need to teach uqi (if I haven't already) how to call JSON.parse on the string coming back when it is a JSON type.

hktouw:
> Since mysql does support a JSON type and uqi supports a JSON type we just need to teach uqi (if I haven't already) how to call JSON.parse on the string coming back when it is a JSON type.

Updated the code to not call JSON.parse when the return value is a JSON type and it is already an array or object. This seemed to address the issue with the double parse. https://github.com/open-truss/open-truss/pull/183/commits/94919ce621bd767ed8e3250918f1aec8fced2b00

> However, it does not seem that uqi knows how to return array types as I'm getting this following error. I think once we teach uqi to know how to take array results from the database and convert them into json arrays we should be good to go.

Adding this `array` type mapping seemed to just fixed it for trino. If we run into this issue again for other clients, we can give this a try there too. https://github.com/open-truss/open-truss/pull/183/commits/e1d20bc3f395be001d978a6fb51167e3d8926854

While I was in here, I took a look at the dynamic signals code and saw it needed some refactoring. Also I noticed it didn't support multidimensional arrays so I thought I'd add support for that. https://github.com/open-truss/open-truss/pull/183/commits/599683875b87488e4560adaf84ea25ba0b70528c. And there is an alternate syntax for this as well. https://github.com/open-truss/open-truss/pull/183/commits/5cb1d2e7271e3e5594347565ce0170560c62e45a

‎localhost:3000otplayground15-arrays-signals-from-query 2024-05-20 at 2 42 07 PM

hktouw (approved):

packages/open-truss/src/signals/index.tsx
```diff
@@ -120,41 +120,40 @@ function signalNameFromObject(signal: object | object[]): string {
   return CryptoJS.SHA256(s).toString()
 }
 
-function createZodShape(signal: object | object[]): {
+type DefaultValue = object | object[] | string | boolean | number
```
kmcq, line 123: What about string[] (and number and bool)?

jonmagic: I'm guessing we will need these but I don't have a workflow right now to exercise them so I'm going to :ship: this as is and we can update once we have a workflow example to exercise/show off the new functionality.

hktouw: Dynamic signals actually already supports arrays for strings, booleans, and numbers. As part of the [recursion for arrays], it grabs the first element in the array and uses that as the element type, then returns a default value array using that element (e.g. `['']`). I guess for whatever reason the types are happy. Perhaps these return values are conforming to `object[]`?

A couple other thoughts:
- This code only uses the type of the first element. This seemed fine since I believe databases only support 1 type (plus null) for a given column, so for most of our current use cases one type is fine. 
- We may eventually need to support union types for signals, e.g. `string | number`. While this isn't necessary for databases, I can see rest APIs or even display components making use of union types for signals. This probably isn't too hard to do with a bit of string parsing, but just noting that we'll need to add support for this.

SLAWEK41 (approved):

PR #186, opened 2024-05-22
demo-app uqi client improvements

jonmagic:
## Why?

The uqi clients in the demo-app could use some improvements based on what I've learned using it at the day job.

## How?

Mostly just copied over changes from the day job.

We're getting close to the point where the uqi clients can be pulled out into separate npm packages. I just want to let things bake another couple of weeks before doing that though.

### DIFF
--- a/demo-app/src/lib/trino-uqi-client.ts
+++ b/demo-app/src/lib/trino-uqi-client.ts
@@ -4,6 +4,7 @@ import {
   type UqiContext,
   type UqiMappedType,
   type UqiResult,
+  type UqiScalar,
 } from '@open-truss/open-truss'
 import {
   BasicAuth,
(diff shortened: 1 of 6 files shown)


demo-app/src/lib/trino-uqi-client.ts
```diff
@@ -73,6 +85,94 @@ async function createTrinoUqiClient(config: TrinoConfig): Promise<UqiClient> {
...
 
+type TrinoColumnValue =
+  | string
+  | number
+  | boolean
+  | object
+  | null
+  | TrinoColumnValue[]
+
+function trinoColumnParser(
```
kmcq, line 96: The ol trino type string parser!

kmcq (approved):
All seems legit to me. Thanks for doing it commit by commit

demo-app/src/lib/trino-uqi-client.ts
hktouw: Can't escape it lol

hktouw (approved):

PR #188, opened 2024-05-23
use object-hash instead of crypto-js for object hashing

kmcq:
## Why?

This is kinda a silly one, but I think still good to do. We use `crypto-js` to perform SHA hashing on a signal object, which is used to give each signal a unique name based on its definition. This package [has a non-ASCII character in its source code](https://github.com/brix/crypto-js/blob/ac34a5a584337b33a2e567f50d96819a96ac44bf/src/ripemd160.js#L2), which is not allowed in my project's build 😅 

## How?

To keep this functionality, which I think we should, I imported and used a different package: [`object-hash`](https://github.com/puleos/object-hash). It does less than `crypto-js`, but that's fine as all we need is the hashing capabilities. (I didn't check, but I'd assume the bundle size decreased due to this change :tada:)

### Testing

I used example 6-state-management-demo to ensure signals are still working.

image

### Considerations

I considered removing this functionality and instead doing something simpler like appending a counter or the current timestamp. However, it does seem nice to give repeated definitions a pointer to the same signal so I kept it. We can revisit if we want of course.

### DIFF
--- a/demo-app/package-lock.json
+++ b/demo-app/package-lock.json
@@ -55,7 +55,7 @@
     },
     "../packages/open-truss": {
       "name": "@open-truss/open-truss",
-      "version": "0.20.0",
+      "version": "0.21.0",
       "license": "MIT",
       "dependencies": {
         "@preact/signals-react": "^2.0.0",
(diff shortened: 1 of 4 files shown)


jonmagic (approved):
🤘

PR #189, opened 2024-05-23
More uqi improvements

jonmagic:
## Why?

Found a few more things that could be improved based on usage at the day job.

## How?

1. Added logic to handle azure mysql connections where user must follow `<user>@<subdomain>` format.
2. Configured teardown function for each uqi client.
3. Removed engine_version property from create client function because we're not using it yet.


demo-app/src/lib/mysql-uqi-client.ts
```diff
@@ -103,13 +103,22 @@ async function createMysqlUqiClient(config: MysqlConfig): Promise<UqiClient> {
     return asyncGenerator()
   }
 
+  let user = parsedUrl.username
```
hktouw, line 106: What do you think about extracting this into a function? `extractUser(url: URL): string`?

hktouw (approved):
This looks great to me 🤘

demo-app/src/lib/mysql-uqi-client.ts
```diff
@@ -136,6 +128,19 @@ async function createMysqlUqiClient(config: MysqlConfig): Promise<UqiClient> {
   })
 }
 
+function azureUserFromUri(parsedUri: URL): string {
```
hktouw, line 131: The way this is structured isn't exclusive to azure per se, so I don't think this needs to be named `azureUserFromUri`. But not a big deal either way.

jonmagic: The regular expression it uses is azure specific 🤷🏻‍♂️

Issue #170, opened 2024-05-09
Support ability to pass in variables into workflows


hktouw:
Currently there is no way to pass in data into a workflow. Data in the workflow needs to be hardcoded into the workflow config or fetched by data provider components. We'll want to come up with an elegant way to do this. Perhaps we can expose a new parameter in `RenderConfig` called `signals` that the caller can supply. We then make the signals available to components in the engine.

/cc @open-truss/engineers for your thoughts on this
/cc @kmcq @pourash-b as you things you are working on may need this functionality

kmcq:
@hktouw yes! This will really help.

A RenderConfig param makes sense to me 👍 I was also thinking that it could be cool to have a config annotation that populates a signal from a query param. That would just be a helper around your idea though probably so let's do yours first.

PR #196, opened 2024-06-10
Pass variables into workflows

fionaochs:
Resolves https://github.com/open-truss/open-truss/issues/170

Created a new variable of `initializedSignals` that is passing a url param through `RenderConfig`  to display an initial value on a new `16-state-management-demo-initial-value` page

I added a new input field on the workflows page where you can initialize an `accountID` that will be used in the  `16-state-management-demo-initial-value` workflow
image

Demo successfully grabs `initialValue` from url and populates the `AccountID` input
image


packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -114,6 +125,11 @@ function createSignals(signalsConfig: SignalsV1, validate: boolean): Signals {
       const s = signal.parse(undefined)
```
kmcq, line 125: I think this might be where you want to use the initial value (instead of undefined) instead of where you have it a few lines down. Parsing undefined gets the default value but in the case when a value's been passed in we want to call parse on that passed in value

kmcq (commented):
Looking legit! I didn't try it out but I have a hunch about that error you're seeing. 

Also, a nitpick, but what do you think about calling it `initialSignalValues` instead of initializedSignals? Initialized to me sounds like we're passing signals but we're just passing in values. Happy to be overruled on this one 😄

demo-app/src/open-truss/components/DisplayInitialAccountId.tsx
```diff
@@ -0,0 +1,20 @@
...
+  BaseOpenTrussComponentV1PropsShape,
+  type z,
+} from '@open-truss/open-truss'
+import { AccountIDSignal } from '../signals'
+
+export const Props = BaseOpenTrussComponentV1PropsShape.extend({
+  accountId: AccountIDSignal,
+})
+
+// This component shows how Open Truss components can use signals (backed by @preact/signals-react) to share state with other components.
```
jonmagic, line 11:
```suggestion
```

packages/open-truss/src/configuration/engine-v1/RenderConfig.tsx
```diff
@@ -111,9 +121,17 @@ function createSignals(signalsConfig: SignalsV1, validate: boolean): Signals {
...
+        const initialValue = initialSignalValues[name]
+        const s = signal.parse(undefined)
+        s.yamlName = name
+        s.value = initialValue
+        signals[name] = s
+      } else {
+        const s = signal.parse(undefined)
+        s.yamlName = name
+        signals[name] = s
+      }
```
jonmagic, line 134: I think this should also work and is a bit easier to read:

```suggestion
      const s = signal.parse(undefined)
      s.yamlName = name
      signals[name] = s

      if (name in initialSignalValues) {
        s.value = initialSignalValues[name]
      }
```

jonmagic (approved):
I applied my suggestions and bumped the version number and I think this is now ready to :ship: so Imma go ahead and do that.

Issue #190, opened 2024-05-24
Update the UqiStatus interface and implementation to return percentage complete


jonmagic:
The UqiStatus interface currently looks like this:

```ts
export interface UqiStatus {
  completedAt: Date | null
  failedAt: Date | null
  failedReason: string | null
  recordsReturned: number
  startedAt: Date | null
}
```

It does not include a percentage complete metric, only the number of rows returned so far.

I considered adding a metric to indicate the number of rows it will return and then calculate the percentage complete manually but that's not possible across all of the databases we currently support.

- Kusto doesn't expose how many rows it will return but it does have an API indicating percentage complete.
- Mysql does expose how many rows it will return but not a percentage complete metric.
- Trino doesn't expose how many rows it will return but does have an API indicating an estimation of percentage complete.

[source](https://chatgpt.com/share/d893ebbe-87c3-440f-b542-4e4c3822883d)

I think we can update uqi to provide an estimated percentage complete metric in the UqiStatus interface, we'll just have to expose a function for each uqi client to define how it calculates this metric.

PR #201, opened 2024-06-18
Report percentageComplete

tmetz:
🍐 with @jonmagic 

https://github.com/open-truss/open-truss/issues/190

This PR lays some groundwork for reporting the percentage of a query that has completed.  We'd like this to work with Kusto, Trino, and MySQL for Uqi, but haven't been able to find a good metric for Kusto or Trino.  However, this PR adds the `percentageComplete` field that can be used if the data is available.

This also does a quick bugfix for the Kusto demo config to make it valid YAML.

### DIFF
--- a/demo-app/package-lock.json
+++ b/demo-app/package-lock.json
@@ -56,7 +56,7 @@
     },
     "../packages/open-truss": {
       "name": "@open-truss/open-truss",
-      "version": "0.23.0",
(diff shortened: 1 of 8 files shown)


jonmagic (approved):
Looks great! Glad mysql was this simple. We'll see if working on kusto and trino clients is easier in the work environment where the uqi client is used in a background job.

PR #204, opened 2024-06-19
0.25.0 -> 0.26.0, pulling in CVE fixes from #202 and #203

spicycode:
This will get #202 and #203 out for any downstream consumers.


hktouw (approved):
LGTM 👍

PR #215, opened 2024-07-27
Improve ergonomics of signals

hktouw:
## Why

@jonmagic pointed out that the ergonomics of using signals can be improved. Right now there are challenges with making components reactive to signal changes and the mental model can be challenging to design for.

### Initial approach

@jonmagic initially spiked out an approach that wraps every component with a Component wrapper, the component wrapper manages signals (and their reactivity) and injects raw values into the component, and the component simply works with the raw values.

This seemed to be feasible for reading signals, but this removes the ability for components to update signals and trigger updates to other components. i.e. components would only have access to raw values instead of updatable signals.

### Approach in this PR

Reading the [@preact/signals-react](https://www.npmjs.com/package/@preact/signals-react) docs, it seems like the recommended approach is to use the babel transpiler. This statically detects when signals are used and updates the code to be reactive. In lieu of this, it is recommended to use the `useSignals()` hook, which when called will make the entire component reactive and re-render when a signal in the component is updated.

Our existing code uses other hooks to achieve reactivity (e.g. `useSignalEffect`, `useComputed`), but `useSignals()` seems to be much simpler to use and requires less knowledge for how signals works.

I updated a few components to use this pattern and they continued to work as expected and the code was refactored to be a bit more obvious and less fragile.

@jonmagic can you point me to the code we were working on the other day that was having reactivity issues? I want to see if changing all the components in that workflow file to using `useSignals()` would address the issues we were seeing.

### Proposal

I propose that we use `useSignals()` for all OT components and see if this makes reactivity easier.

If we use this approach, I think the mental model becomes easier in that we can think of this the same way we would if we used standard DataProvider wrappers. i.e. DataProvider wrappers sets getters / setters that are passed into components for use. When signals are working as expected, the passed in signal is a getter / setter and the access patterns should be similar to using DataProviders.

I propose we give this a try, see what works and doesn't work, then make changes as necessary.

## Code review

The best way to review this PR is commit by commit with the whitespace removed (i.e. `?w=1`), reading each comment along the way.


packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -12,8 +12,7 @@ import {
   type OpenTrussComponent,
   type OpenTrussComponentExports,
 } from '../RenderConfig'
-import React, { useState } from 'react'
```
kmcq, line 15: Keep the React import plz 🙏  my app requires it

packages/open-truss/src/configuration/engine-v1/Frame.tsx
```diff
@@ -93,16 +92,6 @@ export function Frame(props: FrameContext): JSX.Element {
     })
     processedProps._DEBUG_ = debug
 
-    if (frames === undefined) {
-      if (data) {
-        return (
-          <DataProvider {...processedProps} data={data} component={Component} />
-        )
-      } else {
-        return <Component {...processedProps} />
```
kmcq, line 102: Not sure I understand how we can get rid of this part of this logic. Could you say more?

hktouw: Great question! With this block removed, the code will always return `<Component {...processedProps}>{subframes}</Component>`. When `subframes` is `undefined` because no `frames` have been specified in the annotation, it still renders just fine. i.e. `undefined` seems to be a valid component child.

andrew-dillon (approved):
I really like this approach! Elegant and clean ✨

jonmagic (approved):
Let's give it a try 👍 

Can we bump the version too?

jonmagic (commented):
Realized this introduced a bug in OTUqiDataProvider as the fetchData call needs to happen inside of a useEffect.

packages/open-truss/src/components/OTUqiDataProvider.tsx
jonmagic, line 1: After some testing I confirmed that we should have left this file as is. We want a useSignalEffect here as we don't want to update the signal until the query.value has changed.

hktouw: Ah, my bad for introducing a bug ❤️ 

Can you walk me through exactly what behavior you are seeing and what behavior your are expecting? I don't believe the `OTUqiDataProvider` is updating the signal and it should be only reading from it.

jonmagic: It reads query.value but on line 77 it writes signal.value.

hktouw: Oh I see, thanks for pointing that out. I think I see two differences in the refactor. I'm not sure which would cause the bug you are seeing:
- The old code only runs the `useSignalEffect` block when `query.value` has been updated. The new code does this _and_ if any props are updated which trigger a normal re-render of the component.
- In the new code on line 38 we do an early return if `query.value` is an empty string (i.e. `if (query.value === '') return`). This differs from the old code in that the old code would always return `children` even if the `query.value` is empty. The early exit should be updated to return `children`.

I think we could probably refactor this component to use `useSignals` instead of `useSignalEffect`, but I think conceptually I can see an argument made that `useSignalEffect` makes more sense here. i.e. `useSignals` re-renders the entire component when a signal changes, whereas `useSignalEffect` only reruns the callback. The latter is probably more what `OTUqiDataProvider` is doing since we don't care about rendering. So keeping your fix you made makes sense to me 👍 https://github.com/open-truss/open-truss/pull/220

jonmagic: For posterity here is the error I was getting in a workflow that used OTUqiDataProvider after this switching to `0.30.0`:

```
Uncaught runtime errors:
×
ERROR
OTUqiDataProvider(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
    at reconcileChildFibers (http://localhost:3000/index-7d0fc2857e7444616240.js:143869:23)
    at reconcileChildren (http://localhost:3000/index-7d0fc2857e7444616240.js:146690:28)
    at mountIndeterminateComponent (http://localhost:3000/index-7d0fc2857e7444616240.js:147590:5)
    at beginWork (http://localhost:3000/index-7d0fc2857e7444616240.js:148749:16)
    at HTMLUnknownElement.callCallback (http://localhost:3000/index-7d0fc2857e7444616240.js:133645:14)
    at Object.invokeGuardedCallbackDev (http://localhost:3000/index-7d0fc2857e7444616240.js:133694:16)
    at invokeGuardedCallback (http://localhost:3000/index-7d0fc2857e7444616240.js:133756:31)
    at beginWork$1 (http://localhost:3000/index-7d0fc2857e7444616240.js:153659:7)
    at performUnitOfWork (http://localhost:3000/index-7d0fc2857e7444616240.js:152471:12)
    at workLoopSync (http://localhost:3000/index-7d0fc2857e7444616240.js:152402:5)
```

PR #238, opened 2025-08-12
Upgrade zod to v4 and update schemas for compatibility

jonmagic:
## Why?
- Our codebase was relying on zod v3, which is now deprecated and missing key improvements from newer versions.
- Several downstream schemas and utility functions began failing with newer Node.js/TypeScript versions due to incompatibilities with zod v3.

## How?
- Upgrade all references of zod to v4 in root and affected sub-packages ([`package.json`](package.json), [`packages/open-truss/package.json`](packages/open-truss/package.json)).
- Update all schema definitions and utility usages for compatibility with zod v4, including [`config-schemas.tsx`](packages/open-truss/src/configuration/engine-v1/config-schemas.tsx), [`describe-zod.ts`](packages/open-truss/src/utils/describe-zod.ts), and [`signals/index.tsx`](packages/open-truss/src/signals/index.tsx).
- Adjust data provider components ([`OTGraphqlDataProvider.tsx`](packages/open-truss/src/components/OTGraphqlDataProvider.tsx), [`OTRestDataProvider.tsx`](packages/open-truss/src/components/OTRestDataProvider.tsx)) to account for updated schema typing and inference behavior.
  - commit [b8df480](https://github.com/your-org/your-repo/commit/b8df480)

### DIFF
--- a/demo-app/package-lock.json
+++ b/demo-app/package-lock.json
@@ -57,7 +57,7 @@
     },
     "../packages/open-truss": {
       "name": "@open-truss/open-truss",
-      "version": "0.32.0-beta.1",
+      "version": "0.32.0-beta.2",
       "license": "MIT",
       "dependencies": {
         "@preact/signals-react": "^2.0.0",
--- a/demo-app/package-lock.json
+++ b/demo-app/package-lock.json
@@ -57,7 +57,7 @@
     },
     "../packages/open-truss": {
       "name": "@open-truss/open-truss",
-      "version": "0.32.0-beta.2",
+      "version": "0.32.0-beta.3",
       "license": "MIT",
       "dependencies": {
         "@preact/signals-react": "^2.0.0",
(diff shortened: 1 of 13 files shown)


kmcq:
@jonmagic thanks for this! I'll try to see if it's compatible with Figma's setup this week and if it is approve and merge. If it's not then I might want us to think a bit about version management incase we need to patch the zod 3 version for some reason.

Speaking of versions, what are your thoughts on making this a major semver bump? My thinking is you've got it right to not do that  here because OT is at 0.x but I also wonder if we can say this is at v1 currently (and then this PR is v2) given that we've been using it in production for over a year 😎

kmcq:
Okay yea so we're still working on zod 3->4 but I don't think that needs to stop this PR. I created https://github.com/open-truss/open-truss/tree/zod-3 to freeze the current state of things in case we need to patch the zod 3 version.

It would be great to figure out versioning but this is fine for now to me.

cc @harrison-figma in case you have thoughts!

harrison-figma (approved):
@jonmagic If these changes are working for y'all then I think it's 👍 to go ahead and merge this. I think this does block us from being easily able to update OT and pull in changes, but we haven't had the need in a while and have no plans to so that's fine. Like Keith said, we can figure out versioning later when we get there and maybe by then we'd have upgraded to zod 4 as well.

jonmagic:
I'm in no rush to ship this. We're running on the beta release with no issues so we can be patient while y'all catch up to the future 😛

PR #280, opened 2026-05-20
Move `uqi` to `packages/uqi` and remove it from the demo app

### LINKED ISSUE #276: Pull `uqi` out to its own package (`@open-truss/uqi`)

### LINKED ISSUE #277: Remove `uqi`- and data fetching-related parts out of the demo-app

kmcq:
Fixes #276
Fixes #277 

`uqi` and `open-truss` have been determined to be distinct. Keeping them in the same package and demo-app adds complexity to both systems. @jonmagic suggested pulling `uqi` into a closed-source, GitHub-owned code, but I think it's pretty minimal effort to just pull it out of the main OT package and keep it open. It's cool software to have out in the open :)

This PR:
- creates `packages/uqi` and moves all the relevant source into there
- remove `uqi` from the demo-app. we could keep it but it adds complexity that I'd like to remove to make maintenance more manageable
- deletes the OTUqiDataProvider and DataProvider `open-truss/components`. they aren't being used and this removes complexity
- Bumps both packages to version 1.0.0! This is a breaking change so this felt right but also both of these packages are used in production so I think it's appropriate to call them out as stable 🎉


kmcq:
Sorry for the quick ready->draft->ready toggle. I was thinking that I might want to add removing the `Data` key of OT's `Frame` config spec completely now that UQI is out but I think we can do that at a later time

packages/uqi/README.md
```diff
@@ -96,7 +103,9 @@ const client = await createClient({
   username: process.env.KUSTO_USERNAME,
   password: process.env.KUSTO_PASSWORD,
 })
-const queryIterator = await client.query('acme.production.users | project first_name')
+const queryIterator = await client.query(
```
hktouw: seems like mr robot has code style opinions

hktouw (approved):
It's a shame to not have the uqi demo app code available, but it's in the git history so we can bring it back in the future if need be.

Looks good to me, and the project looks much cleaner

packages/uqi/README.md
kmcq: hm yea I'll undo these

kmcq:
@jonmagic The most recent four commits address the feedback you sent me. Thanks for doing that! It was straight forward to address but I'll call out two things:

> One release-strategy question: is the jump to @open-truss/open-truss@1.0.0 meant as a deliberate stability signal for the whole framework API, or mostly version-syncing with the new UQI package? If the API contract is not meant to be stable yet, a 0.33.0 breaking release might communicate the risk more clearly.

I think my PR OP addresses this. It's to indicate a breaking change _and_ stability

> The package tests/examples import UQI through relative source paths, which is fine for unit coverage, but it does not validate the package export path. It would be good for CI or a package-level smoke test to build @open-truss/uqi and import from @open-truss/uqi before publishing.

I didn't test this but I plan on doing so after I merge this. Since consumers of OT are using `^0.##.##` there shouldn't be any issue with merging this as they won't pull in the `1.0.0` release

PR #288, opened 2026-06-09
Fix publishing

kmcq:
The `JS-DevTools/npm-publish` action's README suggests not using it 😅 In this PR I've replaced it with standard `npm publish` commands. Each package step checks the published version before publishing (avoids noisy failures on no-op pushes) and uses `--access public --ignore-scripts` since these are OSS packages.

I also added `docs/PUBLISHING.md` explaining the automated workflow, prerequisites (`NPM_TOKEN` secret), and manual publish instructions.

I also ran `npm pkg fix` on the root and `packages/open-truss` `package.json` files to normalize formatting. (This was a `warn` log in the failed publish action)


.github/workflows/publish-package.yml
```diff
@@ -19,17 +19,33 @@ jobs:
...
         working-directory: ./packages/open-truss
         run: npm run build
-      - uses: JS-DevTools/npm-publish@v3
-        with:
-          token: ${{ secrets.NPM_TOKEN }}
-          package: ./packages/open-truss
+      - name: Publish @open-truss/open-truss
+        working-directory: ./packages/open-truss
+        run: |
+          PUBLISHED=$(npm view . version 2>/dev/null || echo "")
```
jonmagic: Nice move getting us off the action. One thing I think may need a small adjustment. `npm view . version` reads the `latest` dist-tag, not whether this exact local version has already been published. That can still fail on prereleases or anything published without `latest`. Could we check the package name plus `CURRENT` directly, like `npm view "$name@$CURRENT" version`, and publish only when that lookup misses?

docs/PUBLISHING.md
```diff
@@ -0,0 +1,52 @@
...
+## Prerequisites
+
+- An npm [automation token][] with publish access to the `@open-truss` organization must be stored as the `NPM_TOKEN` repository secret.
+
+## Manual publish
+
+To publish a new version locally:
+
+```bash
+npm version <major|minor|patch>
```
jonmagic, line 38: I think this instruction may point at the wrong `package.json`. The workflow publishes from `packages/open-truss` and `packages/uqi`, so a root `npm version` bump would not change the versions the workflow reads. Could we make this say to run `npm version` from the package directory, or add a release script if the root is meant to own that?

package.json
```diff
@@ -24,5 +24,7 @@
   "dependencies": {
     "@open-truss/open-truss": "^0.3.0"
   },
-  "ot:setup": "open-truss setup"
+  "ot:setup": "open-truss setup",
+  "name": "",
```
jonmagic: This may be fine as an `npm pkg fix` side effect, but I think the empty root `name` and `version` make the release path easier to misread. If the root package is not meant to be published, would `private: true` and no root version be clearer?

jonmagic (approved):
Thanks for doing this. I like getting off the publish action and making the workflow and docs more explicit.

I’m approving because the direction looks right to me, but I left a few line comments that I think may need changes before we rely on this for publishing. The main things are making the version guard check the exact version, and making the manual release docs point at the package manifests the workflow actually publishes.

package.json
kmcq: Yea great callout. I wonder if it would be better to move away from having this top-level package.json file and instead have a collection of scripts that `cd` into the various directories (the two packages, demo-app) and run their own lint scripts. I created https://github.com/open-truss/open-truss/issues/299 for this

kmcq:
Thanks for the review @jonmagic! I implemented your feedback in the latest commits

PR #292, opened 2026-06-25
Add docs for OT package and organize config spec versions into their own files

### LINKED ISSUE #291: Write docs for the OT react package
Currently there are none! We should add some docs explaining what the package is, how it's used, and how to develop it.

kmcq:
Closes https://github.com/open-truss/open-truss/issues/291.

We don't have docs and we should! This PR:
- Creates a `config-spec/` dir and a file for each version. This felt more organized to me than the current approach of a single file with a running list
- Makes the root `config-spec.md` a symlink to the `config-spec/v1.0.0.md` file
- Speaking of, I added a `config-spec/v1.0.0.md` file that describes the state of things. This is pretty similar to the `v0.0.2` spec
- Adds a `packages/open-truss/README.md` that is the package's docs


config-spec/v1.0.0.md
```diff
@@ -0,0 +1,87 @@
...
+  version: 1                 # required, positive integer
+  id: my-workflow            # optional, used for cursor persistence
+  debug: false               # optional, enables debug mode
+  signals:                   # optional
+    signalName: signalType   # e.g. integer: number
+  frameWrapper: MyWrapper    # optional component name
+  renderFrames:              # optional
+    type: all                # or: inSequence
+    next: :nextSignal        # required for inSequence
+    back: :backSignal        # optional, for inSequence
```
jonmagic: I could be missing something, but I think the current schema requires `back` for `inSequence`, because `RenderFramesShape` has `back: z.string()`. This calls it optional, and the example below omits it, so I think a copied example will fail default validation unless the schema changes or the docs make `back` required.

packages/open-truss/README.md
```diff
@@ -0,0 +1,359 @@
...
+
+## FrameWrapper
+
+You can provide a custom `FrameWrapper` component to wrap every rendered frame. This is useful for adding UI chrome like delete buttons, drag handles, or navigation controls ([the open-truss demo app's config builder does this](https://github.com/open-truss/open-truss/blob/main/demo-app/src/pages/ot/config-builder/index.tsx)):
+
+```tsx
+export const Props = BaseOpenTrussComponentV1PropsShape.extend({
+  ...withChildren,
+  configPath: z.string(),
+  frame: FrameV1Shape,
```
jonmagic: I think this example may not work for package consumers right now because `FrameV1Shape` is not exported from the public entry point. Maybe the fix is to export it intentionally, or maybe the docs should use the exported `FrameType`/`FrameWrapper` types instead, but as written I think an external user cannot copy this.

jonmagic (approved):
Thanks for putting this together, this is already helpful. I am approving because this is docs-only and the direction makes sense to me, but I left a couple comments where I think the examples may not line up with the current schema/package API. I could be wrong, but they might be worth addressing before folks build on these docs.

kmcq:
Thanks for the review, @jonmagic! I addressed your two comments by updating the docs to say that `back` is required (which it is) and to provide a better FrameWrapper example -- we don't need that export because we don't need to do what that example was doing 😄

